PR #2, opened 2015-02-18
Add support for reading and writing unsigned images (VTK, Nifti).

clangguth:
Ok, let the discussion begin :-)


clangguth:
Please merge https://github.com/unibas-gravis/scalismo/pull/3 BEFORE this one.

clangguth:
@marcelluethi @Ghazi-Bouabene 
From my point of view, this is almost ready for merging. Please review the implementation -- particularly the deprecated methods. We need to find a way to remove them without breaking functionality.

src/main/java/scalismo/image/BSplineCoefficients.java
```diff
@@ -76,7 +76,9 @@ static float getInitialCausalCoefficientMirrorOnBounds(float[] c, float z,
 	} /* end getInitialCausalCoefficientMirrorOnBounds */
 
 
-	
+    // TODO: not actually deprecated, but should be renamed to something that makes it clear that
+    // TODO: the method modifies data in place.
+    @Deprecated
 	static public void getSplineInterpolationCoefficients(int degree, float[] c) throws Exception {
```
marcelluethi: I think this is not really an issue. It is anyway obvious by the type signature (returns void). 
I would argue for just removing the deprecated and make it private (it was never intended for public use).

src/main/scala/scalismo/common/Scalar.scala
```diff
@@ -0,0 +1,283 @@
+package scalismo.common
```
marcelluethi: The file header is missing. 
Are parts of it copied from spire? If yes, we should mention it

src/main/scala/scalismo/common/Scalar.scala
```diff
@@ -0,0 +1,283 @@
...
+      }
+    }
+
+    override def createArray(data: Array[Long]): ValueClassScalarArray[ULong, Long] = ValueClassScalarArray(data)(this)
+    override def toUnderlying(u: ULong): Long = u.toLong
+    override def fromUnderlying(p: Long): ULong = ULong(p)
+  }
+}
+
+sealed trait ScalarArray[S] {
```
marcelluethi: API documentation is missing

src/main/scala/scalismo/common/Scalar.scala
```diff
@@ -0,0 +1,283 @@
...
+  final lazy val size = length
+
+  final def isDefinedAt(index: Int): Boolean = index < size && index >= 0
+
+  def map[T: Scalar: ClassTag](f: S => T): ScalarArray[T]
+
+  def iterator: Iterator[S]
+}
+
+abstract case class AbstractScalarArray[S, U](protected[scalismo] val rawData: Array[U]) extends ScalarArray[S] {
```
marcelluethi: Is this class intended to be used from outside? If not we should make it private. If yes, we should document it.

src/main/scala/scalismo/common/Scalar.scala
```diff
@@ -0,0 +1,283 @@
...
+    val toScalar = implicitly[Scalar[T]]
+    toScalar match {
+      case s: PrimitiveScalar[T] => s.createArray(rawData.map { u => f(fromUnderlying(u)) })
+      case s: ValueClassScalar[T, _] => s.convertArray[U](rawData, { u => f(fromUnderlying(u)) })
+    }
+  }
+
+}
+
+final class PrimitiveScalarArray[S <: AnyVal: ClassTag](rawData: Array[S]) extends AbstractScalarArray[S, S](rawData) {
```
marcelluethi: ditto

src/main/scala/scalismo/common/Scalar.scala
```diff
@@ -0,0 +1,283 @@
...
+  override def equals(that: Any) = {
+    that match {
+      case a: PrimitiveScalarArray[_] => (this canEqual that) && this.rawData.deep == a.rawData.deep
+      case _ => false
+    }
+  }
+
+}
+
+final class ValueClassScalarArray[S <: AnyVal, U <: AnyVal](rawData: Array[U])(implicit scalar: ValueClassScalar[S, U]) extends AbstractScalarArray[S, U](rawData) {
```
marcelluethi: ditto

src/main/scala/scalismo/common/Scalar.scala
```diff
@@ -0,0 +1,283 @@
...
+
+}
+
+object ValueClassScalarArray {
+  def apply[S <: AnyVal, U <: AnyVal](array: Array[U])(implicit s: ValueClassScalar[S, U]): ValueClassScalarArray[S, U] = new ValueClassScalarArray[S, U](array)(s)
+}
+
+object ScalarArray {
+
+  def apply[T: Scalar: ClassTag](array: Array[T]): ScalarArray[T] = {
```
marcelluethi: Would it make sense to allow for the creation of ScalarArrays from arbitrary traversables? 
+ more scalaish
- it is slower, and hence should be discouraged

src/main/scala/scalismo/mesh/TriangleMesh.scala
```diff
@@ -37,7 +36,7 @@ case class TriangleCell(ptId1: Int, ptId2: Int, ptId3: Int) extends Cell {
  * Triangle meshes are currently the only supported representation of 3-dimensional meshes in the library.
  *
  */
-case class TriangleMesh private[scalismo] (private val meshPoints: IndexedSeq[Point[_3D]], val cells: IndexedSeq[TriangleCell], private val cellMapOpt: Option[mutable.HashMap[Int, Seq[TriangleCell]]])
```
marcelluethi: What is the reason for disallowing access to the cell array?

src/main/scala/scalismo/common/Scalar.scala
clangguth: Fixed missing header. No copying from spire :-)

clangguth:
So... any more comments? Just a note of caution: this will introduce some breaking changes when merged.

PR #11, opened 2015-03-16
added statistical model metrics : specificity and generalization

Ghazi-Bouabene:
implementation of specificity and generalization metrics on Statistical mesh models.

### DIFF
--- a/src/main/scala/scalismo/statisticalmodel/dataset/ModelMetrics.scala
+++ b/src/main/scala/scalismo/statisticalmodel/dataset/ModelMetrics.scala
@@ -47,7 +47,6 @@ object ModelMetrics {
    *
    * @param pcaModel Statistical Mesh Model to be evaluated
    * @param dc test data collection that is in correspondence with the model reference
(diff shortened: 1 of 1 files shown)


src/main/scala/scalismo/statisticalmodel/dataset/ModelMetrics.scala
```diff
@@ -0,0 +1,73 @@
...
+      data.map { m => MeshMetrics.avgDistance(m, sample) }.min
+    }.sum * (1.0 / nbSamples)
+  }
+
+  /**
+   * Returns the generalization metric of the Statistical Mesh Model, that is how well the model can represent unseen data
+   *
+   * @param pcaModel Statistical Mesh Model to be evaluated
+   * @param dc test data collection that is in correspondence with the model reference
+   * @param globalUncertainty point uncertainty to be used for the projection. Default is 0.2 mm
```
marcelluethi: do we need this parameter?

PR #18, opened 2015-03-23
nearest neighbor interpolation for DiscreteVectorField

Ghazi-Bouabene:
Not sure if this should also be done for DiscreteScalarField as well. But that might be redundant to the DiscreteScalarImage interpolation. What do you think ?

### DIFF
--- a/src/main/scala/scalismo/common/DiscreteField.scala
+++ b/src/main/scala/scalismo/common/DiscreteField.scala
@@ -35,6 +35,7 @@ trait DiscreteField[D <: Dim, A] extends PartialFunction[Int, A] { self =>
 
   def foreach(f: A => Unit): Unit = values.foreach(f)
 
+  def interpolateNearestNeighbor(): Field[D, A]
   // TODO conceptually, we should have a map here too, but it becomes tricky to
   // do since the overloaded functions will all require their own version of map
   // Maybe a trick with CanBuildFrom and Builder, similar to the scala collectiosn would be required.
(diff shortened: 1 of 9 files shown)


src/main/scala/scalismo/common/DiscreteField.scala
```diff
@@ -78,6 +81,15 @@ class DiscreteVectorField[D <: Dim, DO <: Dim] private (val domain: DiscreteDoma
   override def apply(ptId: Int) = data(ptId)
   override def isDefinedAt(ptId: Int) = data.isDefinedAt(ptId)
 
+  
+  /**
+   * Returns a continuous vector field, where the value at each point is that of the closest point in the discrete set 
+   * **/
+  def interpolateNearestNeighbor()(implicit ev: NDSpace[D], cb: CanBound[D]): VectorField[D, DO] = {
+    val indexedDomain = SpatiallyIndexedDiscreteDomain(domain.points.toIndexedSeq, domain.numberOfPoints)(ev)
```
marcelluethi: For computational reasons, we might want to do this only if the domain is not yet SpatiallyIndexed (we can find this out by pattern matching). This would make it possible to apply it also to e.g. Triangle Meshes, without incurring twice the costs of indexing

marcelluethi:
We can have it on the DiscreteField definition, and thus have it automatically on vector and scalar fields (and images). For the images we need to make sure that we override (calling the b-spline interpolation of 0-th order), in order to avoid that a huge structure is indexed.

Ghazi-Bouabene:
One thing still to be discussed is that when calling interpolateNearestNeighbor for DiscreteScalarImage, this would compute the kd-tree for the grid. Overriding this method to use interpolate(0) does not immediately work as the types do not match

marcelluethi:
@Ghazi-Bouabene You are right, it makes sense to have them in the constructor.
However, we need to rename (or rework) the typeclasses. With the current names, their meaning is not clear at all (especially when they are put in a more global context). I will propose something after having merged it.

PR #20, opened 2015-03-24
Changed vector field domain to real space at the low rank Gaussian

ghost:
Changed vector field domain to real space at the low rank Gaussian process interpolation.


marcelluethi:
I agree in principle. Choosing the full RealSpace as a domain makes much more sense than the bounding box.

There is a small glitch in the PR. In Line 207, you still use the boundingBox.

PR #29, opened 2015-04-08
Introduced DiscreteGaussianProcess and generalized marginalization idea

marcelluethi:
Instead of only a LowRankDiscreteGaussianProcess, we now also have a DiscreteGaussianProcess.
Furthermore, marginalization for GaussianProcesses has been extended such that any finite number
of points can be used for marginalizing a distribution.


clangguth:
Hooray! We're slowly getting there!

http://geek-and-poke.com/geekandpoke/2014/1/2/games-for-the-real-geeks-part-2

(SCNR, don't take this too seriously)

PR #35, opened 2015-04-17
Introduce explicit type and case class for KLBasis

marcelluethi:
KLBasis used to be defined as Seq[(Float, VectorField)]. We introduced a type alias
for the seq and a case class for the tuple (Float, VectorField). The changes were made for
DiscreteLowRankGaussianProcess and LowRankGaussianProcess.

Before merging this pull request, we should discuss the terminology, as these things will be all visible to the user and somehow define how the user will think about things. Any comments?


src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
```diff
@@ -222,6 +227,9 @@ case class DiscreteLowRankGaussianProcess[D <: Dim: NDSpace: CanBound, DO <: Dim
 
 object DiscreteLowRankGaussianProcess {
 
+  case class Eigenpair[D <: Dim, DO <: Dim](eigenvalue: Float, eigenfunction: DiscreteVectorField[D, DO])
```
Ghazi-Bouabene: rename to DiscreteEigenPair ?

PR #39, opened 2015-04-30
Trait CoordinateOps to abstract common functionality for Point, Vector and Index

marcelluethi:
Functionality like map, ... was implemented in all the subclasses of Coordinate. 
As in the future, it is likely that we will add more methods, such as filter, zip, ... we should avoid this code duplication. This PR is an attempt to abstract functionality into a common trait.

### DIFF
--- a/src/main/scala/scalismo/geometry/Coordinate.scala
+++ b/src/main/scala/scalismo/geometry/Coordinate.scala
@@ -17,6 +17,8 @@ package scalismo.geometry
 
 import breeze.linalg.DenseVector
 
+import scala.reflect.ClassTag
+
 /**
  * The basic n-tuple in R^n^ with scalar type S
  */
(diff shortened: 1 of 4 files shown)


src/main/scala/scalismo/geometry/Coordinate.scala
```diff
@@ -41,3 +43,24 @@ abstract class Coordinate[D <: Dim: NDSpace, @specialized(Int, Float, Double) S]
   override def toString = data.deep.toString()
 }
 
+trait CoordinateOps[D <: Dim, Scalar, Repr <: Coordinate[D, Scalar]] { self: Coordinate[D, Scalar] =>
+
+  implicit val classTagScalar: ClassTag[Scalar]
+
+  protected def createRepr(data: Array[Scalar]): Repr
```
clangguth: suggestion: createRepresentation()

src/main/scala/scalismo/geometry/Point.scala
```diff
@@ -16,11 +16,15 @@
...
 import scala.language.implicitConversions
+import scala.reflect.ClassTag
 
 /**
  * An n-dimensional Point
  */
-class Point[D <: Dim: NDSpace] private (private[scalismo] override val data: Array[Float]) extends Coordinate[D, Float] {
+class Point[D <: Dim: NDSpace] private (private[scalismo] override val data: Array[Float]) extends Coordinate[D, Float] with CoordinateOps[D, Float, Point[D]] {
+
+  override val classTagScalar = implicitly[ClassTag[Float]]
```
clangguth: explicitly state type. :-p

src/main/scala/scalismo/geometry/Index.scala
```diff
@@ -16,9 +16,15 @@
 package scalismo.geometry
 
 import scala.language.implicitConversions
+import scala.reflect.ClassTag
+
+class Index[D <: Dim: NDSpace] private (val data: Array[Int]) extends Coordinate[D, Int] with CoordinateOps[D, Int, Index[D]] {
+  val classTagScalar: ClassTag[Int] = implicitly[ClassTag[Int]]
```
clangguth: explicitly state that this is an override

src/main/scala/scalismo/geometry/Coordinate.scala
```diff
@@ -41,3 +43,24 @@ abstract class Coordinate[D <: Dim: NDSpace, @specialized(Int, Float, Double) S]
...
+
+  def mapWithIndex(f: (Scalar, Int) => Scalar): Repr = {
```
clangguth: check if utils.ArrayUtils.fastMap can be used, the array mapping is already implemented there.

PR #43, opened 2015-05-01
Implement gaussian smoothing of discrete images (using vtk)

marcelluethi:
Implementation of Gaussian smoothing for DiscreteScalarImages. Although it could in theory be implemented by using the convolution method of ScalarImage, we use the VTK implementation for performance reasons..


marcelluethi:
@Ghazi-Bouabene I think a typeclass would be overkill here. The pattern matching is merely used to decide whether we need to call the vtk method that takes 2 arguments or the one that takes 3. Even if we would at one point introduce 4D, this code would remain the same, since VTK only support 2D and 3D..

Ghazi-Bouabene:
OK. But for the sake of argument, this case is very similar to rotation, defined for 2 and 3D, but not for 1D. And there, we all agreed on a type-class based solution !?

Ghazi-Bouabene:
On the other hand, I just realized that the the 1D case is handled by the CanConvertToVTK implicit. So it's fine by me as is.

PR #49, opened 2015-05-23
Added functionality to treat ASM as a distribution :

Ghazi-Bouabene:
- introduced DiscreteFeatureField class : field from point to arbitrary DenseVector[Float]
- introduced ASM Sample class. These are the instances sampled from the ASM : a mesh + a DiscreteFeatureField at the profile points
- added mean and sample methods to the ActiveShapeModel class
- added sampleFeaturesOnly to the ActiveShapeModel class. This is a utility method to be used for visualization/debugging

### DIFF
--- a/src/main/scala/scalismo/statisticalmodel/asm/ActiveShapeModel.scala
+++ b/src/main/scala/scalismo/statisticalmodel/asm/ActiveShapeModel.scala
@@ -1,15 +1,15 @@
 package scalismo.statisticalmodel.asm
 
 import breeze.linalg.DenseVector
-import scalismo.common.SpatiallyIndexedDiscreteDomain
+import scalismo.common._
 import scalismo.geometry.{ Point, _3D }
 import scalismo.image.DiscreteScalarImage
 import scalismo.mesh.TriangleMesh
-import scalismo.numerics.{ Sampler, FixedPointsUniformMeshSampler3D }
+import scalismo.numerics.Sampler
 import scalismo.registration.{ LandmarkRegistration, RigidTransformation, Transformation }
 import scalismo.statisticalmodel.{ MultivariateNormalDistribution, StatisticalMeshModel }
 
-import scala.collection.{ AbstractIterator, immutable }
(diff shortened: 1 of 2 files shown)


src/main/scala/scalismo/statisticalmodel/asm/ActiveShapeModel.scala
```diff
@@ -91,6 +98,39 @@ case class ActiveShapeModel(statisticalModel: StatisticalMeshModel, profiles: Pr
...
+    val featureField = DiscreteFeatureField(randomProfilePoints zip randomFeatures)
+    ASMSample(sampleMesh, featureField)
+  }
+
+  /**
+   * Utility function that allows to randomly sample different feature profiles, while keeping the profile points
+   * Meant to allow to easily inspect/debug the feature distribution
+   */
+
+  def sampleFeaturesOnly: ASMSample = {
```
clangguth: This should explicitly be defined as sampleFeaturesOnly() -- i.e., with parentheses. In contrast, the mean() method *could* (I'm not saying "should", I don't know what's better) be defined in this accessor-like style without parentheses.

src/main/scala/scalismo/statisticalmodel/asm/Profiles.scala
```diff
@@ -17,9 +18,31 @@ case class Profiles(domain: SpatiallyIndexedDiscreteDomain[_3D], data: immutable
...
 }
 
+/**
+ * Class of Discrete Fields where to each point, a DenseVector of arbitrary dimensionality is associated.
+ *
+ * An example instance of such a class is the set of profile points and associated features in an Active Shape Model.
+ *
+ */
+
+case class DiscreteFeatureField[D <: Dim: NDSpace](private val pointsAndValues: IndexedSeq[(Point[D], DenseVector[Float])]) extends DiscreteField[D, DenseVector[Float]] {
```
clangguth: Should this stay in the asm subpackage, or should it go into the "common" package where the other fields are defined?

PR #56, opened 2015-06-08
Typeclasses cleanup and domain refactoring

marcelluethi:
Introduce again explicit subclasses for all concrete instances of type {1D,2D,3D}. This makes the implementation more straight-forward and reduces typeclass related clutter. 

Along the way we had to refactor the domain concept, as the class hierarchy as it was did not allow for these changes.


marcelluethi:
@Ghazi-Bouabene  Thanks for the review. I implemented your proposals for improvements and added an additional unit test.

PR #61, opened 2015-06-14
Massive performance increase for ASM sampling

clangguth:


src/main/scala/scalismo/statisticalmodel/asm/ActiveShapeModel.scala
```diff
@@ -80,10 +80,11 @@ case class ActiveShapeModel(statisticalModel: StatisticalMeshModel, profiles: Pr
    * Returns the mean mesh of the shape model, along with the mean feature profiles at the profile points
    */
   def mean(): ASMSample = {
-    val meanProfilePoints = profiles.data.map(p => statisticalModel.mean.point(p.pointId))
+    val smean = statisticalModel.mean
+    val meanProfilePoints = profiles.data.map(p => smean.point(p.pointId))
     val meanFeatures = profiles.data.map(_.distribution.mean)
     val featureField = DiscreteFeatureField(new UnstructuredPointsDomain3D(meanProfilePoints), meanFeatures)
-    ASMSample(statisticalModel.mean, featureField, featureExtractor)
```
Ghazi-Bouabene: after looking at statisticalModel.mean I Understand why the improvement is so "massive" :). But wouldn't a better fix be then to make the mean a val instead of a def in StatisticalModel ?

PR #66, opened 2015-07-08
Introduce neighborhood concept in discrete domain

marcelluethi:
Introduced the concept of neighbors of a point in a domain. 
The concept and interface is kept extremely simple. 

@sandroschoenborn @Ghazi-Bouabene , do you think that is sufficient for a start? Do you think we need something more complicated?


sandroschoenborn:
I would also vote for a separate GraphDomain trait. In general, I think the concept of neighborhood is important but only to some applications, so moving it to a graph domain would not obscure it too much.

marcelluethi:
The concept of graph domain will be implemented once the new mesh structures are integrated and finalized. The implementation will be as a separate trait, as suggested by @Ghazi-Bouabene  and @sandroschoenborn

PR #70, opened 2015-07-21
Separate cases VTK_SIGNED_CHAR and VTK_CHAR in vtk conversion. Fixes #69

### LINKED ISSUE #69: Conversion from Signed Char image fails with ClassCastException
The following code
```
 case VTK_CHAR | VTK_SIGNED_CHAR =>
         val in = arrayVTK.asInstanceOf[vtkCharArray].GetJavaArray()
```
results in a classCastException as a```vtkSignedCharArray``` cannot be cast into 
a ```vtkCharArray```. A solution would be to treat the cases ```VTK_CHAR``` and ```VTK_SIGNED_CHAR``` separately.

marcelluethi:


clangguth:
Nice catch, thanks!

PR #71, opened 2015-07-21
Point/Vector/Index Refactoring with specialized classes

sandroschoenborn:
A refactoring of the basic types `Point`/`Vector`/`Index` based on specific implementations for each dimensionality.

## Motivation:

* `Coordinate` hierarchy boxes
* slow beyond boxing due to the Array overhead, also in specific contexts (fixed dimensionality)
* access through `apply(i: Int)` led to unreadable code

## Features

* no more boxing, no generic Arrays
* specific types `Vector1D`, `Vector2D` and `Vector3D` with named members: usage e.g. `vec.x`
* all specific access and creation operations are faster (25% for rendering)
* preserves all generic operations in the form of `map` and `mapWithIndex`
* implicit conversions between general types and specific implementations: `Index[_3D]` and `Index3D`
* includes many new tests (in `GeometryTests`)

## Downsides

* generic apply access `apply(i: Int)` operations with a slight overhead due to indirect member access
* some duplicated code

## Technical Details

* storage backend with named members, no more Arrays
* generic creation needs a typeclass
* no "naked" data views anymore, any access as Array yields a copy (`point.data` is marked *deprecated* and linked to `toArray`, replaced by `point.toArray`)

### Create Typeclasses

Each of the classes provides a `Create[D]` typeclass to allow for generic construction. These would quickly clutter all generic code since they are needed to construct from Arrays, a very widespread operation. To hide these typeclasses, they have have been mixed into the `NDSpace[D]` object which is already present everywhere. So, no change is necessary to continue using old code like `Vector[D](array)`

### DIFF
--- a/src/main/scala/scalismo/common/DiscreteField.scala
+++ b/src/main/scala/scalismo/common/DiscreteField.scala
@@ -117,11 +117,11 @@ object DiscreteVectorField {
    * the vector is ordered as (v_11, v_12, ... v_1d, ...v_n1, v_n2, v_nd)
    */
   def fromDenseVector[D <: Dim: NDSpace, DO <: Dim: NDSpace](domain: DiscreteDomain[D],
-    vec: DenseVector[Float]): DiscreteVectorField[D, DO] = {
+    vec: DenseVector[Float])(implicit vectorBuilder: Vector.Create[DO]): DiscreteVectorField[D, DO] = {
     val dim = implicitly[NDSpace[DO]].dimensionality
     val vectors =
       for (v <- vec.toArray.grouped(dim))
-        yield Vector[DO](v)
+        yield Vector.apply[DO](v)
 
     DiscreteVectorField[D, DO](domain, vectors.toIndexedSeq)
   }
--- a/src/main/scala/scalismo/common/DiscreteField.scala
+++ b/src/main/scala/scalismo/common/DiscreteField.scala
@@ -117,7 +117,7 @@ object DiscreteVectorField {
    * the vector is ordered as (v_11, v_12, ... v_1d, ...v_n1, v_n2, v_nd)
    */
   def fromDenseVector[D <: Dim: NDSpace, DO <: Dim: NDSpace](domain: DiscreteDomain[D],
-    vec: DenseVector[Float])(implicit vectorBuilder: Vector.Create[DO]): DiscreteVectorField[D, DO] = {
+    vec: DenseVector[Float]): DiscreteVectorField[D, DO] = {
     val dim = implicitly[NDSpace[DO]].dimensionality
     val vectors =
       for (v <- vec.toArray.grouped(dim))
--- a/src/main/scala/scalismo/geometry/Index.scala
+++ b/src/main/scala/scalismo/geometry/Index.scala
@@ -28,7 +28,7 @@ sealed abstract class Index[D <: Dim: NDSpace] {
 
   def toArray: Array[Int]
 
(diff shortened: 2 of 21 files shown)


src/main/scala/scalismo/geometry/Index.scala
```diff
@@ -15,30 +15,123 @@
...
+    override def zero: Index[D] = zeros[D]
+    override def plus(x: Index[D], y: Index[D]): Index[D] = x.mapWithIndex((v, i) => v + y(i))
+  }
 
   object implicits {
-    implicit def index1DToInt(v: Index[_1D]) = v(0)
-    implicit def intToindex1De(i: Int): Index[_1D] = Index(i)
-    implicit def tupleOfIntToindex2D(t: (Int, Int)): Index[_2D] = Index(t._1, t._2)
-    implicit def tupleOfIntToindex3D(t: (Int, Int, Int)): Index[_3D] = Index(t._1, t._2, t._3)
+    implicit def point1DToInt(ind: Index[_1D]): Int = ind.i
```
clangguth: nitpicking: should be named index1DToInt

src/main/scala/scalismo/io/StatismoIO.scala
```diff
@@ -221,7 +221,7 @@ object StatismoIO {
   private def writeRepresenterStatismov090(h5file: HDF5File, group: Group, model: StatisticalMeshModel, modelPath: String): Try[Unit] = {
 
     val cellArray = model.referenceMesh.cells.map(_.ptId1) ++ model.referenceMesh.cells.map(_.ptId2) ++ model.referenceMesh.cells.map(_.ptId3)
-    val pts = model.referenceMesh.points.toIndexedSeq.par.map(p => (p.data(0).toDouble, p.data(1).toDouble, p.data(2).toDouble))
+    val pts = model.referenceMesh.points.toIndexedSeq.par.map(p => (p.toArray(0).toDouble, p.toArray(1).toDouble, p.toArray(2).toDouble))
```
clangguth: p => (p.x.toDouble, p.y.toDouble, p.z.toDouble) ?

clangguth:
Looks great to me!

src/main/scala/scalismo/common/DiscreteField.scala
```diff
@@ -117,11 +117,11 @@ object DiscreteVectorField {
    * the vector is ordered as (v_11, v_12, ... v_1d, ...v_n1, v_n2, v_nd)
    */
   def fromDenseVector[D <: Dim: NDSpace, DO <: Dim: NDSpace](domain: DiscreteDomain[D],
-    vec: DenseVector[Float]): DiscreteVectorField[D, DO] = {
+    vec: DenseVector[Float])(implicit vectorBuilder: Vector.Create[DO]): DiscreteVectorField[D, DO] = {
```
marcelluethi: Is there a reason we cannot create the vector using NDSpace, but need to pass the Vector.Create?

marcelluethi:
@sandroschoenborn  Looks great! Thanks a lot for the changes. 
Maybe now that we store the data for points as Floats (x,y,z), we could get rid of the array altogether and use an immutable seq.  This would make it possible to use data again, but this time in a safe way. Anyway, this is a separate issue that we can tackle in the future.

src/main/scala/scalismo/common/DiscreteField.scala
sandroschoenborn: no, this is old relict, can now be replaced by NDSpace[D]

PR #72, opened 2015-07-24
Added valueType to ASM preprocessor, made FeatureExtractor also work with intensity images

clangguth:


marcelluethi:
@clangguth Thanks for the changes. It's obviously an improvement of the current state and I will merge it as is.  However, I have the feeling we need further discussions for refinements of that mechanism.

PR #73, opened 2015-07-30
update to sbt version 0.13.7

marcelluethi:
Since typesafe [migrated their repos](https://www.typesafe.com/blog/incident-report-for-repo-typesafe-com-and-repo-scala-sbt-org)  using ```sbt 0.13.0``` leads to an obscure compilation error. Upgrading to a more recent sbt version resolves the problem.


clangguth:
Thanks! :-)

PR #74, opened 2015-08-10
Introduce explicit type PointId (instead of Int) in DiscreteDomain.

marcelluethi:
So far PointIds were simply represented as Ints. In this PR we introduce an explicit type (value class) PointId. This makes method signatures more clear and avoids confusion when there are several different types of Ids, such as for example in ASMs, where we have to distinguish between PointId and ProfileId. 

Besides introducing the type PointId, we also introduce a type ProfileId in the ASM classes.

### DIFF
--- a/src/main/scala/scalismo/common/DiscreteDomain.scala
+++ b/src/main/scala/scalismo/common/DiscreteDomain.scala
@@ -16,30 +16,35 @@
 package scalismo.common
 
 import scalismo.geometry._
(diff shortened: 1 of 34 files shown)


Ghazi-Bouabene:
I agree. This reduces confusion especially in the case of active shape models.
This was also a good occasion to learn more about value-classes :)  (to really leverage them we just have to make sure never to stack the domain points into an array to avoid allocation)

PR #76, opened 2015-08-18
ImageIO cleanup

clangguth:
- removed support for .h5 image files, und thus for 1D images
- some minor code cleanup and optimization


marcelluethi:
@clangguth  Thanks - looks good. It is great to see that we are sometimes also deleting some code.

PR #78, opened 2015-08-19
Readability: fixed some obvious typos, rephrased some messages, remov…

clangguth:
…ed some comments.

No functional changes.


clangguth:
Note: I've created few more branches, namely cleanup_cosmetic, cleanup_arrays, and cleanup_redundant_val (in that order), which are all diffs to the "previous" one. Ultimately, IMO they should all be merged, but I'm not sure what the correct way of merging such incremental changes is. So I'm only creating a PR for the first part of the changes... and then we'll discuss and see. :-)

marcelluethi:
@clangguth Thanks a lot for taking this initiative.

PR #81, opened 2015-08-25
Added Identity as a convenience method to RigidTransformation

clangguth:


marcelluethi:
This is sorely needed. However, it should not be addressed only at the level of the rigid transform, but for all transformations. As this is no urgent request, I will close it and work on a solution for all transformations.

PR #84, opened 2015-09-02
Image resampling speedup

Ghazi-Bouabene:
Multiple optimizations to speedup image resampling.

For a reasonably sized image (332 * 365 * 393), the speedup of resampling a cubic spline interpolation on the same domain is 10x (9.7). 

As a repercussion, image alignment with a cubic spline interpolation has a speedup of 16x.
(An alignment with level-0 interpolation had a speedup of 35x)


marcelluethi:
Awesome - thanks a lot!!  Just had a few minor comments.
Maybe we should at the end squash everything into one commit.

PR #90, opened 2015-10-13
findClosestPoint returns case class instead of Tuple

Ghazi-Bouabene:
One new case class. The rest are mechanical changes


clangguth:
+1 :-)

marcelluethi:
:+1:

PR #99, opened 2016-02-12
Marginal likelihood

Ghazi-Bouabene:


marcelluethi:
@Ghazi-Bouabene  Awesome. This is really useful to have.

PR #101, opened 2016-03-16
Added ImageIO methods with implicit voxel type conversion

clangguth:


src/main/scala/scalismo/io/ImageIO.scala
```diff
@@ -218,6 +218,54 @@ object ImageIO {
...
+   * This method is similar to the [[read3DScalarImage]] method, except that it will convert the image to the requested voxel type if
+   * the type in the file is different, whereas [[read3DScalarImage]] will throw an exception in that case.
+   *
+   * @param file  image file to be read
+   * @param resampleOblique  flag to resample oblique images. This is only required when reading Nifti files containing an oblique image. See documentation above [[ImageIO]].
+   * @param favourQform  flag to favour the qform Nifti header entry over the sform one (which is by default favoured). See documentation above [[ImageIO]].
+   * @tparam S Voxel type of the image
+   *
+   */
+  def read3DScalarImageAsType[S: Scalar: TypeTag: ClassTag](file: File, resampleOblique: Boolean = false, favourQform: Boolean = false): Try[DiscreteScalarImage[_3D, S]] = {
```
Ghazi-Bouabene: Couldn't this be implemented as a flag castIfneeded defaulted to false (as the resampleOblique), instead of defining a new reading function ?

clangguth: Good idea. The only problem with this kind of "flag options" is that they can become tedious and unmaintainable over time, especially if there are also default values and/or if people rely on the order. But in principle, yes, and I'm not against it.

Ghazi-Bouabene: I acutally meant to reply to THIS mail in the previous reply I sent you :)


  On 17.03.2016 13:45, Christoph Langguth wrote:
>
> In src/main/scala/scalismo/io/ImageIO.scala 
> <https://github.com/unibas-gravis/scalismo/pull/101#discussion_r56497344>:
>
> > @@ -218,6 +218,54 @@ object ImageIO {
> >      }
> >    }
> >
> > +  /**
> > +   * Read a 3D Scalar Image, and possibly convert it to the requested voxel type.
> > +   *
> > +   * This method is similar to the [[read3DScalarImage]] method, except that it will convert the image to the requested voxel type if
> > +   * the type in the file is different, whereas [[read3DScalarImage]] will throw an exception in that case.
> > +   *
> > +   * @param file  image file to be read
> > +   * @param resampleOblique  flag to resample oblique images. This is only required when reading Nifti files containing an oblique image. See documentation above [[ImageIO]].
> > +   * @param favourQform  flag to favour the qform Nifti header entry over the sform one (which is by default favoured). See documentation above [[ImageIO]].
> > +   * @tparam S Voxel type of the image
> > +   *
> > +   */
> > +  def read3DScalarImageAsType[S: Scalar: TypeTag: ClassTag](file: File, resampleOblique: Boolean = false, favourQform: Boolean = false): Try[DiscreteScalarImage[_3D, S]] = {
>
> Good idea. The only problem with this kind of "flag options" is that 
> they can become tedious and unmaintainable over time, especially if 
> there are also default values and/or if people rely on the order. But 
> in principle, yes, and I'm not against it.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub 
> <https://github.com/unibas-gravis/scalismo/pull/101/files/d3cf5f2a7f552511f426b8c835d9056a048985de#r56497344>
>


-- 
Ghazi Bouabene, PhD
Department of Mathematics and Computer Science
University of Basel
+41 61 267 0538

marcelluethi:
I am in favor of Christoph's proposal, as it makes it more explicit in the code that a casting happens (compared to simple a true/false in the argument list). I still consider the casting of image types dangerous, and therefore think it is good to be explicit about it.

clangguth:
I guess we can talk about this in person tomorrow, so in any case, don't merge it just yet. If only because I made a copy/paste mistake, and the 2D version has useless arguments ;)

PR #102, opened 2016-03-22
ScalarmeshField read

Ghazi-Bouabene:
Implemented method to read ScalarMeshFields. Main point to discuss: 
To keep life simple, I read the data to Float by default. Objections?


Ghazi-Bouabene:
Just updated the PR to add support for different Scalar types (as orally discussed).

PR #103, opened 2016-04-05
Point & Vector with additional functionality: normalize, multiplyComponents

sandroschoenborn:
Added a few methods to `Point` and `Vector`, namely:

- Vector: `.normalize` returns normalized version of vector
- Vector: `.multiplyComponents` multiplies component-wise
- constants: `origin` in Point, `unit` in Vector for unit length base vectors, `zero` in Vector for zero vectors
- convenience constructors from polar and spherical coordinates for Point
- Interface types changed to generic, e.g. `Point[_3D]` instead of `Point3D` where applicable

existing functionality should be unaffected


marcelluethi:
Thanks a lot for adding this functionality. 
I have only two comments:
1) A unit test for the polar/spherical constructors would be nice
2) The multiplyComponents could be written as :*, similar to how it is defined in breeze

PR #105, opened 2016-04-05
Replaced `val dimensionality` by `def` -> stable initialization

sandroschoenborn:
in `Vector`/`Point`/`IntVector` the field `dimensionality` was a `val` which led to troubles with initializer blocks in depending code (is not properly initialized and has value `0`)

- replaced by `def dimensionality` in `Vector`/`Point`/`IntVector`


marcelluethi:
@sandroschoenborn Good catch! Thanks!

PR #109, opened 2016-04-20
Pivoted Cholesky Approximation

thogerig:
* Pivoted Cholesky Approximation
* RandomSVD in kernel class replaced by PC-approximation
* Statistical mesh model has a new method to augment a model.


src/main/scala/scalismo/statisticalmodel/StatisticalMeshModel.scala
```diff
@@ -215,8 +216,12 @@ object StatisticalMeshModel {
   }
 
   /**
-   *  Adds a bias model to the given statistical shape model
+   *  @deprecated
```
marcelluethi: It would be helpful to add a short message, which the users sees during compilation
```@deprecated("some reason","20-04-2016")```

marcelluethi:
@thogerig This pull request conflates the improved augmenting of the model with the new approximation. Both functionalities are great, but I suggest to have them as separate commits (and maybe even separate pull requests. This will make it easier to understand what has happened.

PR #113, opened 2016-04-21
Using doubles internally

Andreas-Forster:
The type for real values is now internally consistently `Double` except for memory critical 3D images. To stay compatible with the existing file formats only floats will be written and read. This should not introduce any change in behavior except:

- The methods work with a different precision and therefore only lead nearly the same results.
- A model built, written to a file and loaded again is only equal up to the different precision of Floats and Doubles.

An example for a slightly different numerical result is the ProcrustesDistanceMetric test. Internally the estimated rotation matrix is exactly the identity matrix using float precision. Using Doubles the rotation is only very close but not identical to an identity matrix. The test had to be changed to allow a small error in the range of 1.e-10 .


marcelluethi:
This looks good and first tests showed that everything works as before. 
As these changes affect almost all files, and it is not possible to spot all difficulties, I suggest we merge but test it thoroughly in the next release candidate.

PR #114, opened 2016-04-25
Point in Triangle test in 3d

Andreas-Forster:
Added helper function `pointInTriangle3D` to the object `BarycentricCoordinates` that tests if a 3d point is whithin a 3d triangle.

Nothing else changed.


Andreas-Forster:
Should not be merged! Should go into develop not master

PR #116, opened 2016-05-04
Generic Value Typed Gaussian Processes

Andreas-Forster:
General Value Type:
Introduced Gaussian Processes with general type `Value` as output instead of `Vector[DO]`.
OLD: `GaussianProcess[D, DO].apply(Point[D]): Vector[DO]`
NEW: `GaussianProcess[D, Value].apply(Point[D]): Value`

Vectorizer:
The introduced change needed the concept of a `Vectorizer`. The vectorizer is passed as impiclit parameter. The functionality should remain the same where the implicit `VectorVectorizer` of the `Vector` companion object can be found. The `DiscreteField` companion object offers methods to vectorizer and unvectorize a DiscreteField or an IndexedSeq[Value].

Kernels:
The kernels return now simply a `DenseMatrix[Double]`. They have an additional field `outputDim` instead of the dimension `[_ DO]`.

UseCase:
As an example the `scalismo.statisticalmodel.StatisticalMeshModel` is a working example where a Gaussian Processes from `Point[_3D]` to `Vector[_3D]` is used.


src/main/scala/scalismo/common/Vectorizer.scala
```diff
@@ -0,0 +1,15 @@
+package scalismo.common
```
marcelluethi: here the header is missing

src/main/scala/scalismo/kernels/Kernel.scala
```diff
@@ -78,78 +84,111 @@ abstract class MatrixValuedPDKernel[D <: Dim: NDSpace, DO <: Dim: NDSpace] { sel
...
+    DiscreteMatrixValuedPDKernel[D](domain, k, outputDim)
   }
 
 }
+
 @deprecated("use DiagonalKernel instead", "0.10.0")
-case class UncorrelatedKernel[D <: Dim: NDSpace](kernel: PDKernel[D]) extends MatrixValuedPDKernel[D, D] {
-  val I = SquareMatrix.eye[D]
-  def k(x: Point[D], y: Point[D]) = I * (kernel(x, y)) // k is scalar valued
+case class UncorrelatedKernel[D <: Dim: NDSpace](kernel: PDKernel[D], override val outputDim: Int) extends MatrixValuedPDKernel[D] {
```
marcelluethi: We should use this refactoring to delete the UncorrelatedKernel

src/test/scala/scalismo/statisticalmodel/GenericGaussianRandomFieldTests.scala
```diff
@@ -0,0 +1,52 @@
+package scalismo.statisticalmodel
```
marcelluethi: This file should not be commited, or turned into tests (without the printlns)

src/main/scala/scalismo/common/Vectorizer.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import breeze.linalg.DenseVector
+
+/**
+ * Created by forand00 on 03.05.16.
+ */
+trait Vectorizer[Value] {
+  def dim: Int
```
marcelluethi: should we use size or length instead of dim?

src/main/scala/scalismo/common/Vectorizer.scala
```diff
@@ -0,0 +1,15 @@
+package scalismo.common
+
```
marcelluethi: The file header is missing

marcelluethi:
@Andreas-Forster Thanks a lot for. This looks excellent to me (except the two very minor changes mentioned above). 

I am looking forward to use it to build my first intensity model with scalismo.

src/main/scala/scalismo/common/Vectorizer.scala
Andreas-Forster: i think length or size would suggest that it is related to the vertorizer itself. But instead it refers to the dimension (size/length) of the dense vector representing the value type. Therefore i would argue for to leave it as dim.

src/main/scala/scalismo/common/Vectorizer.scala
Andreas-Forster: updated

src/test/scala/scalismo/statisticalmodel/GenericGaussianRandomFieldTests.scala
Andreas-Forster: removed file

src/main/scala/scalismo/common/Vectorizer.scala
Andreas-Forster: updated

src/main/scala/scalismo/kernels/DiscreteMatrixValuedPDKernel.scala
```diff
@@ -25,13 +25,13 @@ import scalismo.geometry.{ NDSpace, Dim, SquareMatrix }
  *  Mathematically, it can be represented as a covariance matrix. However, it has more structure, i.e. its entry ij
  *  is a matrix. Furthermore, the class has the knowledge about its domain (the point on which it is defined).
  */
-class DiscreteMatrixValuedPDKernel[D <: Dim: NDSpace, DO <: Dim: NDSpace] private[scalismo] (val domain: DiscreteDomain[D],
-    val k: (PointId, PointId) => SquareMatrix[DO]) {
+class DiscreteMatrixValuedPDKernel[D <: Dim: NDSpace] private[scalismo] (
+    val domain: DiscreteDomain[D],
+    val k: (PointId, PointId) => DenseMatrix[Double],
+    val outputDim: Int) {
```
Ghazi-Bouabene: By having the outputdim as a field, we lose the type safety where we could check that the output dimentionalities matched. Do you think it would make sense to also rely on an implicit vectorizer here (even if we don't use it), to figure out the output dimensionality from the vectorizer. The compiler might have more chances to help out in case of mixing kernels with different output dimensionality.

Ghazi-Bouabene:
I also agree on the general structure of the changes. This will add a lot of potential applications with GPs. 
Thanks a lot!

marcelluethi:
As the only possible objection seems to be that we are loosing the type safety in the output dimension (for which there is currently no obvious fix), I am going to merge the pull request in order not to block further developments. As soon as we have a fix to make the output dimension type-safe, we can always fix that "problem" in  a separate PR.

PR #118, opened 2016-05-30
HDF5Utils issue fix: Various Try[Unit] methods always succeeded

sandroschoenborn:
Overview
-------------

- Fixed the `Try[Unit].map`/`flatMap` issue in `HDF5Utils`: The methods returned `Success(())` even on failure:
    - `writeNDArray`
    - `writeInt`
    - `writeFloat`
    - `writeString`

- Added a test which tries to write an unknown data type as `NDArray`, it should result in an exception
- Included possible failing of `h5file.get(path)` in return `Try` (method could still throw)

Details
---------

In `HDF5Utils` Methods with `Try[Unit]` as return type have an issue: they use `map` where `flatMap` should be used.

A problem usually caught by the type system. Not with `Try[Unit]`. When using `map`, the return type of the anonymous function is inferred to be of type `Unit` - compatible with `Success` and `Failure` - the type system is happy, the programmer not so much.

```scala
    val a: Try[Unit] = Success(1.0).map{v => Success()}
    println(s"map Success: $a") // map Success: Success(())

    val b: Try[Unit] = Success(1.0).map{v => Failure(new Exception("fail"))}
    println(s"map Failure: $b") // map Failure: Success(())

    val c: Try[Unit] = Success(1.0).flatMap{v => Success()}
    println(s"flatMap Success: $c") // flatMap Success: Success(())

    val d: Try[Unit] = Success(1.0).flatMap{v => Failure(new Exception("fail"))}
    println(s"flatMap Failure: $d") // flatMap Failure: Failure(java.lang.Exception: fail)
```

### DIFF
--- a/src/main/scala/scalismo/io/HDF5Utils.scala
+++ b/src/main/scala/scalismo/io/HDF5Utils.scala
@@ -85,7 +85,7 @@ class HDF5File(h5file: FileFormat) extends Closeable {
     }
   }
 
-  def writeIntAttribute(path: String, attrName: String, attrValue: Int) = {
+  def writeIntAttribute(path: String, attrName: String, attrValue: Int): Try[Unit] = {
     Try {
       val s = h5file.get(path)
       val fileFormat: FileFormat = FileFormat.getFileFormat(FileFormat.FILE_TYPE_HDF5)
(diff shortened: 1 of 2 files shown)


marcelluethi:
@sandroschoenborn Good catch, thanks. I will also add the fix to the latest release/master

PR #120, opened 2016-05-30
variance parameter for `coefficients`

Andreas-Forster:
Pulled out variance parameter for `coefficients` method in `DiscreteLowRankGausianProcess`. Using the default values results in the exact same behavior. The parameter is necessary to allow the use of estimated noise parameters.


marcelluethi:
The methods coefficients and project need to be consistent. Either none of them have a noise parameter or both. 

@sandroschoenborn  Would it maybe make sense to set it in the constructor, giving it a similar interpretation to the noise in ProbabilisticPCA?

sandroschoenborn:
Added the parameter `sigma2` (variance of independent Gaussian noise per point in observed data) to all `project` and `coefficients` where it was missing

Open point: `logpdf`/`pdf` how to handle model and observation?

Issue #83, opened 2015-08-26
parameters method in RotationTransform should be removed


Ghazi-Bouabene:
This is due to the fact that a Rotation could have been created from a Rotation space around any given center, it does not make sense to return parameters as one would not know on which Rotation Space to use them.

marcelluethi:
solved with PR #123. We are now exposing the center, which is sufficient to interpret the parameters.

PR #123, opened 2016-06-07
Rigid transform with public rotation (and rotation center) and translation

marcelluethi:
So far there is no way to access the rotation center used in a rigid transformation. This makes the parameters of a rotation meaningless (cf Issue #83). This fix resolve this issue.

Furthermore, by exposing rotation and translations in the Rigid Transform we make it possible to access their parameters individually.


Ghazi-Bouabene:
Looks good :) Thanks

PR #124, opened 2016-06-15
bugfix: creating mesh with empty cell list failed

Ghazi-Bouabene:
minBy throws exception on empty list


Andreas-Forster:
Looks fine to me. I think this could be merged as soon as it is checked by Travis CI. Thx for fixing this.

PR #126, opened 2016-06-24
MeshBoundary class

Andreas-Forster:
Added two a mesh boundary interface with implementation for TriangularMesh. The class provides function to check if points, lines or triangles lie on the mesh boundary given their indices. The behaviour is:

- A line connection two points contained in only one triangle lies on the mesh boundary.
- A line connection two points contained in either none or two triangles lies on the boundary.

- Both points of a line which lies on the boundary, are part of the boundary.

- Triangles with one of the sides lying on the boundary is part of the boundary.
- A triangle with only one point on the boundary is not part of the boundary.

Tests are provided and may help to get started.

### DIFF
--- a/src/main/scala/scalismo/mesh/MeshBoundary.scala
+++ b/src/main/scala/scalismo/mesh/MeshBoundary.scala
@@ -1,0 +1,142 @@
+/*
+ * Copyright 2015 University of Basel, Graphics and Vision Research Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package scalismo.mesh
+
+import breeze.linalg.CSCMatrix
+import scalismo.geometry.Dim
+
+
+/**
+  * MeshBoundary is a property to test if points or a line between two points is on the boundary of the mesh.
+  */
+trait MeshBoundary {
+
+  /**
+    * Check if the point with given index is on the boundary of the underlaying mesh.
+    *
+    * @return True if point is part of the boundary.
+    */
+  def pointIsOnBoundary(id: Int): Boolean
+
+  /**
+    * Check if the line between the two points with given indices is on the boundary of the underlaying mesh.
+    * @note This method does not check weather there exists a line between the two points. You will get false for all
+    *       lines not in the mesh.
+    *
+    * @return True if the line is part of the boundary.
+    */
+  def lineIsOnBoundary(id1: Int, id2: Int): Boolean
+
+}
+
+
+
+/**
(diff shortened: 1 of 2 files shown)


src/main/scala/scalismo/mesh/MeshBoundary.scala
```diff
@@ -0,0 +1,142 @@
...
+  def pointIsOnBoundary(id: Int): Boolean
+
+  /**
+    * Check if the line between the two points with given indices is on the boundary of the underlaying mesh.
+    * @note This method does not check weather there exists a line between the two points. You will get false for all
+    *       lines not in the mesh.
+    *
+    * @return True if the line is part of the boundary.
+    */
+  def lineIsOnBoundary(id1: Int, id2: Int): Boolean
```
sschoenborn: I think it should better be called `edgeIsOnBoundary`, you really use *edges* between vertices. A line does not need to lie between two vertices.

src/main/scala/scalismo/mesh/MeshBoundary.scala
```diff
@@ -0,0 +1,142 @@
...
+private class BoundaryOfATriangleMesh(private var vertexIsOnBorder: IndexedSeq[Boolean],
+                                      private var lineIsOnBorder: CSCMatrix[Boolean], // NOTE: CSCMatrix is a yet experimental sparse matrix in breeze
+                                      private var triangleIsOnBorder: IndexedSeq[Boolean]) extends TriangularMeshBoundary {
+
+  override def pointIsOnBoundary(id: Int) : Boolean = {
+    return vertexIsOnBorder(id)
+  }
+
+  override def lineIsOnBoundary(id1: Int, id2: Int): Boolean = {
+    return lineIsOnBorder(id1,id2);
```
sschoenborn: the operation `CSCMatrix(row, col)` is not necessarily very fast for large columns (does a binary search in the column)

maybe consider a `Map[(PointId, PointId), Boolean]` and ensure symmetry through ordering access?

src/main/scala/scalismo/mesh/MeshBoundary.scala
```diff
@@ -0,0 +1,142 @@
...
+    val lineOnBorder = new CSCMatrix.Builder[Boolean](nPts,nPts).result()
+    val triangleOnBorder = new Array[Boolean](nTriangles)
+
+    // lines in only a single triangle lie on the border,
+    // lines contained in two triangles are not on a border
+    triangles.map{ triangle =>
+      linesOfATriangle.map{ line =>
+        val v1 = triangle.pointIds(line._1).id
+        val v2 = triangle.pointIds(line._2).id
+        lineOnBorder(v1,v2) = ! lineOnBorder(v1,v2)
```
sschoenborn: growing a `CSCMatrix` by direct element access is very slow, it constantly reallocates

use:
```
val builder = new CSCMatrix.Builder[Boolean](nPts,nPts)
builder.add(i, j, value)
...
val matrix = builder.result()new CSCMatrix.Builder[Boolean](nPts,nPts)
```

(or change the data structure)

sschoenborn:
I think this is very useful. I just have two comments:

- implementation with `CSCMatrix` seems not optimal for me, maybe use a `Map`? or build the matrix more efficiently
- I think *line* is confusing, rather use *edge* as it always refers to an edge between two vertices, whereas lines can be more general

Andreas-Forster:
Renamed line as suggested to edges.
CSCMatrix is actually faster than using a map for the used test example. During the build the CSCMatrix.Builder is now used and the built matrix is reduced to contain only non-zero entries. This improves the building of the index structure.

sschoenborn:
look good

PR #129, opened 2016-06-27
Interpolator as implicit field in SurfacePointProperty

sschoenborn:
Interpolator as implicit `val`: `implicit val ops: Interpolator[A]`

- easier copying: no own context of `Interpolator[A]` required at copying site since it can be accessed through the `val`

### DIFF
--- a/src/main/scala/scalismo/mesh/MeshSurfaceProperty.scala
+++ b/src/main/scala/scalismo/mesh/MeshSurfaceProperty.scala
@@ -46,7 +46,7 @@ case class ConstantProperty[@specialized(Double, Float, Int, Boolean) A](overrid
 }
 
 /** property per vertex, with interpolation */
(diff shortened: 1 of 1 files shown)


src/main/scala/scalismo/mesh/MeshSurfaceProperty.scala
```diff
@@ -46,7 +46,7 @@ case class ConstantProperty[@specialized(Double, Float, Int, Boolean) A](overrid
 }
 
 /** property per vertex, with interpolation */
-case class SurfacePointProperty[@specialized(Double, Float, Int, Boolean) A](override val triangulation: TriangleList, pointData: PointId => A)(implicit ops: Interpolator[A])
+case class SurfacePointProperty[@specialized(Double, Float, Int, Boolean) A](override val triangulation: TriangleList, pointData: PointId => A)(implicit val ops: Interpolator[A])
```
marcelluethi: @sschoenborn  If we make it public, shouldn't we use a more descriptive name? It is not really clear (at least to me) that Ops refers to the interpolator.

sschoenborn:
renamed the field to `interpolator` which is more descriptive

PR #130, opened 2016-06-27
Bugfix: Handle special case when triangle list is empty

marcelluethi:
Sometimes it is useful to define a mesh that has an empty list of triangles. For example, when we crop a mesh it may happen that one part does not contain any triangles any more. The current implementation crashes in such a case. 

This PR handles this special case.


Ghazi-Bouabene:
This was already addressed in PR (#124) that was merged with branch release-0.12. Closing this PR.

PR #131, opened 2016-06-27
add constructor to define transformation from a function

marcelluethi:
Introduces new factory method (apply method) which takes a function ```Point[D] => Point[D]``` and defines a transformation from it, which is defined on the full real space ```RealSpace[D]```.


Ghazi-Bouabene:
agreed.

PR #132, opened 2016-06-27
Make asm profile a traversable

marcelluethi:
Make profile extend Traversable. This let's us use map, filter and foreach to traverse the profiles, which allows for cleaner and more readable code.


Ghazi-Bouabene:
a welcome change

PR #133, opened 2016-06-27
 Allow construction of DiscreteScalarXXX objects from a Traversable

marcelluethi:
So far only ScalarArrays could be chosen to define DiscreteScalar objects. It is often more convenient (especially for light objects) to use the output of a sequence directly.


Ghazi-Bouabene:
Looks good. Thx.

PR #134, opened 2016-07-01
Closest point on surface

Andreas-Forster:
The new scalismo.mesh.boundingSpheres package offers functionality to find the closest point on a triangulated surface. An example usage is:
`import scalismo.mesh.boundingSpheres._`

`val mesh: TriangleMesh3D = ???`
`val sd = SurfaceDistance.fromTriangleMesh3D(mesh)`

`val p = Point(0,0,0)`
`val (cp,dist2) = sd.getClosestPoint(p)`

There is also the functionality to find the closest point in a set of points. Further the package is open to be extended to find the closest point on a set of lines, eg a future LineMesh.

### DIFF
--- a/src/main/scala/scalismo/mesh/boundingSpheres/PointSetDistance.scala
+++ b/src/main/scala/scalismo/mesh/boundingSpheres/PointSetDistance.scala
@@ -13,33 +13,69 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package scalismo.mesh.surfaceDistance
+package scalismo.mesh.boundingSpheres
 
 import breeze.numerics._
 import BSDistance.{Distance2, Index, _}
 import scalismo.geometry.{Point, Vector, _3D}
+import scalismo.mesh.TriangleMesh
 
 
-case class PointSetDistance(private val bs: BoundingSphere,
-                            private val points: Seq[Point[_3D]]) {
-  val pointList = points.map(_.toVector).toIndexedSeq
-  var lastIdx: Int = 0
+trait PointSetDistance {
+  def closestPoint(point: Point[_3D]): (Point[_3D], Double, Int)
+}
 
 
-  def closestPoint(point: Point[_3D]): (Double, Point[_3D], Int) = {
-    val p = point.toVector
-    val index = Index(lastIdx)
-    val lastP = pointList(lastIdx)
-    val lastD = toPoint(lastP, p)
-    val d: Distance2 = new Distance2(lastD.distance2)
-    distanceToPartition(p, bs, d, index)
-    (d.distance2, pointList(index.idx).toPoint, index.idx)
+/**
+  * PointSetDistance
+  */
+object PointSetDistance {
+
+  /**
+    * Create PointSetDistance for a list of points.
+    */
+  def fromPointList(points: Seq[Point[_3D]]): PointSetDistance = {
+    val bs = BoundingSpheres.createForPoints(points)
+    new PointSetDistanceImplementation(bs, points)
   }
 
-  def distanceToPartition(point: Vector[_3D],
-                          partition: BoundingSphere,
-                          result: Distance2,
-                          index: Index): Unit = {
+  /**
+    * Create PointSetDistance for a list of points.
+    */
+  def fromMesh(mesh: TriangleMesh[_3D] ): PointSetDistance = {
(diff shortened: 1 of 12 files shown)


src/main/scala/scalismo/mesh/surfaceDistance/BSDistance.scala
```diff
@@ -0,0 +1,233 @@
...
+    */
+  @inline
+  private[surfaceDistance] def calculateBarycentricCoordinates(triangle: Triangle, p: Vector[_3D]): (Double, Double, Double) = {
+    val x = triangle.a - p
+    val ab2 = triangle.ab dot triangle.ab
+    val abac = triangle.ab dot triangle.ac
+    val ac2 = triangle.ac dot triangle.ac
+    val xab = x dot triangle.ab
+    val xac = x dot triangle.ac
+    val x2 = x dot x
```
Ghazi-Bouabene: this one is not used :)

src/main/scala/scalismo/mesh/surfaceDistance/SurfaceDistance.scala
```diff
@@ -0,0 +1,269 @@
...
+  * Descritpion of a closest point
+  * @param distance2 the squared distance
+  * @param pt the coordinates
+  * @param ptType on which geometric entity of the surface the closest point lies
+  * @param bc the barycentric coordinates of the point. The interpretation depends on the ptType.
+  * @param idx the index in the original surface instance of the geometric entity where the closest point lies. The interpretation depends on the ptType.
+  */
+case class ClosestPoint(val distance2: Double,
+                        pt: Vector[_3D],
+                        ptType: ClosestPointType,
```
Ghazi-Bouabene: This is a bit too C-ish way of typing  :)  .. If these points really need to be differentiated then why not create real types?

src/main/scala/scalismo/mesh/boundingSpheres/PointSetDistance.scala
```diff
@@ -0,0 +1,145 @@
...
+
+trait PointSetDistance {
+  def closestPoint(point: Point[_3D]): (Point[_3D], Double, Int)
+}
+
+
+/**
+  * PointSetDistance
+  */
+object PointSetDistance {
```
Ghazi-Bouabene: I would be in favor of naming this VantagePointTree instead of PointSetDistance which is anyway confusing as one expects a distance to be returned when calling the factory methods. If we want a generic name instead of the algorithm' one, then I would be in favor of SpatialIndex

src/main/scala/scalismo/mesh/boundingSpheres/PointSetDistance.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+    * Create PointSetDistance for a list of points.
+    */
+  def fromPointList(points: Seq[Point[_3D]]): PointSetDistance = {
```
Ghazi-Bouabene: would be great if it takes a <: Traversable[Point[_3D]]. Avoid annoying compilation errors "expected Seq[..] found List[..]"

src/main/scala/scalismo/mesh/surfaceDistance/SurfaceDistance.scala
Andreas-Forster: Using the C-ish way is simply faster internally. We can use the same mutable result during the search internally. This prevents a lot of class creations during the recursive search. We could of course hide it from the user and return an abstract BaseClosestPointType with only the distance and the point. The user could then pattern match to get the exact point type. However this function will be used often when comparing two meshes where a small peformance difference could be accumulate over a full program. Yet i have no benchmark.

src/main/scala/scalismo/mesh/boundingSpheres/PointSetDistance.scala
Andreas-Forster: Will be updated.

src/main/scala/scalismo/mesh/boundingSpheres/PointSetDistance.scala
Ghazi-Bouabene: OK, I realise now that there are different types of indexes PointSetDistance, SurfaceDistance .. Maybe something like DiscreteSpatialIndex, SurfaceSpatialIndex

PR #135, opened 2016-07-01
Method to convert NDimensionalNormalDistribution to MultivariateNorma…

Ghazi-Bouabene:
…lDistribution

Currently GP.posterior method requires noise as a  MultivariateNormalDistribution, while landmark's uncertainty is an NDimensionalNormalDistribution. Hence the need for conversion.

An alternative could also be to make NDimensionalNormalDistribution a descendant of MultivariateNormalDistribution.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -12,7 +12,7 @@ import sbtbuildinfo.Plugin._
 
 object BuildSettings {
   val buildOrganization = "ch.unibas.cs.gravis"
-  val buildVersion = "0.13.0-RC2"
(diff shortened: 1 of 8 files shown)


Andreas-Forster:
I am not against this pull request. But I suggest to mark the `NDimensionalNormalDistribution` as deprecated and start to replace the class with `MultivariateNormalDistribution`. In the core framework it is only used for the landmarks and could be replaced easily i think. The class could then be thrown out if not a strong resistance from the community arises.

Ghazi-Bouabene:
Updated the PR according to the comment above.

PR #137, opened 2016-07-05
Introduce polylines in 2D and 3D as addition mesh class

marcelluethi:
Initial proposal for adding Polylines in 2D and 3D. 

This PR is meant as a basis for discussions. In particular we should discuss if 

* The names are appropriate for the concepts we want to cover
* All concepts make sense for both 2D and 3D? 
* Whether it fits currently known use cases


src/main/scala/scalismo/mesh/LineContourProperty.scala
```diff
@@ -0,0 +1,76 @@
+package scalismo.mesh
```
Ghazi-Bouabene: Header missing here

src/main/scala/scalismo/mesh/PolyLine.scala
```diff
@@ -0,0 +1,171 @@
...
+ */
+case class LineCell(ptId1: PointId, ptId2: PointId) extends Cell {
+
+  /** Identifiers of the points belonging to the cell*/
+  val pointIds = IndexedSeq(ptId1, ptId2)
+
+  /** Returns true if the given point identifier is part of the triangle cell*/
+  def containsPoint(ptId: PointId) = ptId1 == ptId || ptId2 == ptId
+
+  def toIndex2D = IntVector(ptId1.id, ptId2.id)
```
Ghazi-Bouabene: The purpose of this method is confusing. Plus it does not seem to be used.

src/main/scala/scalismo/mesh/PolyLine.scala
```diff
@@ -0,0 +1,171 @@
...
+}
+
+abstract class PolyLine[D <: Dim: NDSpace](val pointSet: UnstructuredPointsDomain[D], val topology: LineList) {
```
Ghazi-Bouabene: If I understand it correctly, the use of PointIDs is similar to the Triangle mesh, meaning that the point with Id 0 is not necessarily at the beginning of the line. Should there be such a concept (startPoint) here to easily allow for recursive traversals? (Especially given that a cell direction is enforced)

Ghazi-Bouabene:
Looks good. I would vote in favor of calling it LineMesh with goes inline with the TriangleMesh already in place (and the underlying vtk IO calls)

Andreas-Forster:
Before merging a the search results for "triangle" in comments and function names should be checked if they are correct or if they should be changed to "line". Apart of the energed comments i think this looks good.

src/main/scala/scalismo/mesh/PolyLine.scala
marcelluethi: @Ghazi-Bouabene If I understand your proposal correctly, you think about something like a walk in the line. The current proposal does not enforce that the polyline is one connected line. It is therefore difficult to do this in general.

Ghazi-Bouabene: OK I guess the enforceConsistentCellDirections function is the one that made me think of line "walks". Is this meant to be a public method?

sschoenborn:
I think the concept of "porting" the TriangleMesh to a LineMesh is very appealing. Everything we know about the mesh still works very similarly. 2D and 3D make sense to me, lines are meaningful entities in a plane and in 3d space.

regarding the name: `PolyLine` might not be optimal, has this polygon-flair ... `LineMesh` would fit the already existing `TriangleMesh` well, mesh, however, is conceptually a bit tied to the idea of a triangulation - but I think this can be extended to a polygon-type of "triangulation" where the line fits again. 

=> I vote for `LineMesh`

src/main/scala/scalismo/mesh/LineList.scala
```diff
@@ -0,0 +1,114 @@
...
+package scalismo.mesh
+
+import scalismo.common.PointId
+
+final case class LineId(id: Int) extends AnyVal {
+  def isValid: Boolean = this != LineId.invalid
+}
+
+object LineId {
+  val invalid = TriangleId(-1)
```
sschoenborn: should be `LineId(-1)`

src/main/scala/scalismo/mesh/LineList.scala
```diff
@@ -0,0 +1,114 @@
...
+    val mapData = pointMap.mapValues(s => s.toSet) // make immutable
+    val seqData = IndexedSeq.tabulate(pointIds.size) { i => mapData(pointIds(i)).toIndexedSeq }
+    id => seqData(id.id)
+  }
+
+  /**
+   * lines connected to line via common point
+   *
+   */
+  lazy val adjacentTrianglesForTriangle: LineId => IndexedSeq[LineId] = {
```
sschoenborn: should be renamed to  `Line`

src/main/scala/scalismo/mesh/PolyLine.scala
sschoenborn: this is an old relict from internal use with the gravis faces meshes which used an `IntVector` to specify a triangle cell. Not needed anymore (actually also `TriangleCell.toIndex3D` can be removed).

src/main/scala/scalismo/mesh/PolyLine.scala
sschoenborn: also without walks, a consistent ordering (of all individual line segments) is still necessary if you want normals to point in a consistent direction (usually outside)

src/main/scala/scalismo/mesh/PolyLine.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  /** Get all cell normals as a surface property */
+  lazy val cellNormals: LineProperty[Vector2D] = {
+    LineProperty(topology, lines.map(computeCellNormal))
+  }
+
+  /** Get all vertex normals as a surface property */
+
+  lazy val vertexNormals: ContourPointProperty[Vector2D] = {
+
```
sschoenborn: should be `ContourPointProperty[Vector[_2D]]`: `LineContourProperty[A]` is invariant, so no conversion available from  `ContourPointProperty[Vector2D]` to  `ContourPointProperty[Vector[_2D]]`

src/main/scala/scalismo/mesh/PolyLine.scala
```diff
@@ -0,0 +1,171 @@
...
+    val pd = pointSet.pointIds.map { id =>
+      val cellIds = topology.adjacentLinesForPoint(id)
+      val n = cellIds.foldLeft(Vector.zeros[_2D])((acc, cellId) => acc + computeCellNormal(topology.lines(cellId.id)))
+      n / cellIds.size: Vector2D
+    }
+
+    ContourPointProperty(topology, pd.toIndexedSeq)
+  }
+
```
sschoenborn: code duplication? available as `ContourPointProperty.averagedPointProperty` - although less performant, but easier to read

marcelluethi:
Thanks for all the reviews. A cleaned up version is available as PR #140

PR #141, opened 2016-07-07
VantagePointTree: fast search in metric spaces

sschoenborn:
Added a [Vantage-Point Tree](https://en.wikipedia.org/wiki/Vantage-point_tree) for fast neighbour searches in metric spaces. The tree builds a binary space partition using the provided metric (which is exchangeable).

**WARNING:** this only works for a metric (positive, symmetric and triangle inequality), see [Wikipedia](https://en.wikipedia.org/wiki/Metric_(mathematics))

Key concept: at each node store a point and a radius, then split remaining points into an *inner* and *outer* set:

- inner: all points which are closer to the node's point (=pivot) than radius
- outer: all points which are farther away from the pivot than radius

- Added to `scalismo.utils`
- Recursive tree implementation for building and traversal
- includes tests

### DIFF
--- a/src/main/scala/scalismo/utils/VantagePointTree.scala
+++ b/src/main/scala/scalismo/utils/VantagePointTree.scala
@@ -1,0 +1,283 @@
+/*
+ * Copyright 2015 University of Basel, Graphics and Vision Research Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package scalismo.utils
+
+import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


Ghazi-Bouabene:
My understanding is that this is the algorithm behind the recent contribution for fast closest point computation for discrete point sets and surfaces. Are there other known shape modelling use cases where such an algorithm would be needed for a general purpose metric space? In other words, should this be part the Scalismo API as there would be shape modelling needs for it, or simply an internal (private) implementation?

Andreas-Forster:
No this is another algorithm thats used here. This one does only works for triangles or surfaces.

sschoenborn:
It is a different algorithm. The fast `closestPointOfSurface` use a *bounding volume* strategy which is a different thing. The Vantage-point tree is useful for closest-point retrieval in a point set.

Ghazi-Bouabene:
OK, but for discrete point set retrieval, we currently have DiscreteSpatialIndex and we also have SurfaceSpatialIndex for surfaces. Is there a concrete case of a generic metric space where the user would need to perform nearest neighbor queries? Cause otherwise I would be in favor of making it private[scalismo]

sschoenborn:
We like to use it externally (outside package `scalismo`), e.g. for texture synthesis with retrieval based on multiple color values.

Ghazi-Bouabene:
In that case :+1:  :)

marcelluethi:
@Ghazi-Bouabene @sschoenborn  anything more to add here, or is it good merge?

sschoenborn:
Good for me

PR #143, opened 2016-07-14
Default center argument only in one method. Also, ASM fitting updates…

Ghazi-Bouabene:
… pose using the same center

The center can be default"ed" either in rigid3DLandmarkRegistration(IndexedSeq[(Point[_3D], Point[_3D])]) or rigid3DLandmarkRegistration(Seq[Landmark[_3D]],  Seq[Landmark[_3D]]) .. not in both (compiler error). Favored the latter as it is most user friendly.

### DIFF
--- a/src/main/scala/scalismo/io/ImageIO.scala
+++ b/src/main/scala/scalismo/io/ImageIO.scala
@@ -372,7 +372,7 @@ object ImageIO {
       val scaledPS = origPs.map(anisotropicScaling)
       val imgPs = origPs.map(transVoxelToWorld)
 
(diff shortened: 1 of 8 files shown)


marcelluethi:
@Ghazi-Bouabene I think it is confusing to have two methods that do the same, but only one takes a default parameter. In this case i would opt for making center a mandatory argument in both cases.

marcelluethi:
@Ghazi-Bouabene  On second thought: You can replace a method that takes a default parameter by two methods: e.g.
``` 
def foo(a : Int, b: Int = 0) =???
```
becomes 
```
def foo(a : Int) = foo(a, 0)
def foo(a : Int, b : Int = ???
```

Ghazi-Bouabene:
Thanks, updated accordingly.

PR #147, opened 2016-08-02
Simplify MeshSurfaceProperty: buffered data, no specialization

sschoenborn:
Simplification of `MeshSurfaceProperty`:

- Removed type-specialization from all definitions
- Data of most concrete properties is supplied as a `IndexedSeq[A]`
- Added some documentation strings

Especially the buffered data enables a much cleaner case class code, e.g. they have proper equality and hash code methods out of the box.

We lose a tiny bit of flexibility through this: supplying data through the functional `PointId => A` or `TriangleId => A` enabled adhoc solutions without the need the compute a property for each vertex/triangle in advance. But having these functions made a complicated equality test necessary (sample all values first).

If the functionality is really needed now, the way to go would be to define your own `MeshSurfaceProperty` where you need it. It is just a two-methods interface.

### DIFF
--- a/src/main/scala/scalismo/mesh/MeshSurfaceProperty.scala
+++ b/src/main/scala/scalismo/mesh/MeshSurfaceProperty.scala
@@ -20,7 +20,7 @@ import scalismo.common.PointId
 import scala.reflect.ClassTag
 
 /** general property defined on the surface of a mesh */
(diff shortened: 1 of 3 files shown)


marcelluethi:
@sschoenborn Thanks, this simplification makes a lot of sense. The change looks good to me.

PR #149, opened 2016-08-12
extend marginal-likelihood computation to models with non-zero mean

marcelluethi:
The marginal likelihood did not properly handle Gaussian processes with non-zero mean. This is fixed by this PR.


Ghazi-Bouabene:
Looks good. Thanks.

PR #150, opened 2016-08-15
bugfix: vtkconversion of 3D line mesh

marcelluethi:
For the 3D case, the z component of the points that defined the mesh was set to 0 instead
of the proper z value.


marcelluethi:
closed due to erroneous commit (sorry for that). Will be reopened once fixed.

PR #152, opened 2016-08-17
VantagePointTree Controlled Randomness

andreas-schneider:
Until now scala.util.Random was used with changing seed.
With this pull-request it is possible to specify the randomness.
Usage:
implicit val rnd = Random(seed)
val tree = VantagePointTree...


marcelluethi:
Looks good to me. 
@Andreas-Forster If you are also happy with the change, can you merge the PR?

marcelluethi:
@andreas-schneider There is an internal consensus (offline discussion) that the Randomness problem in Scalismo needs to be addressed more fundamentally. The idea is to include a small, dedicated Random class, which provides an implicit instance of a random number generator if the user does not define his/her own implicit instance. 

As this will cover your use case as well (and should be available in develop within the next 2 weeks), I will close this PR.

andreas-schneider:
Great!

PR #154, opened 2016-08-18
renamed SymmetricTransition to SymmetricTransitionRatio

Ghazi-Bouabene:
Currently one can define a proposal with UnitTransition as such:

case class Proposal(..) extends ProposalGenerator[A] with UnitTransition[A] 

while for a symmetric one :

case class Proposal(..) extends ProposalGenerator[A] with TransitionProbability[A] with SymmetricTransition[A]

since SymmetricTransition is in fact a TransitionRatio and not a TransitionProbability. 

This PR hopefully harmonizes the naming.

### DIFF
--- a/src/main/scala/scalismo/sampling/proposals/MixtureProposal.scala
+++ b/src/main/scala/scalismo/sampling/proposals/MixtureProposal.scala
@@ -15,7 +15,7 @@
  */
 package scalismo.sampling.proposals
 
-import scalismo.sampling.{ ProposalGenerator, SymmetricTransitionRatio, TransitionProbability }
+import scalismo.sampling.{ ProposalGenerator, SymmetricTransition, SymmetricTransitionRatio, TransitionProbability }
 
 import scala.util.Random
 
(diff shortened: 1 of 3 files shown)


sschoenborn:
the key change for me is `trait SymmetricTransition[A] extends TransitionProbability[A] with SymmetricTransitionRatio[A]` -- I agree, this should make it easier to choose the right trait

src/main/scala/scalismo/sampling/proposals/MixtureProposal.scala
```diff
@@ -75,10 +75,10 @@ object MixtureProposal {
...
+  def fromSymmetricProposals[A](proposals: (Double, ProposalGenerator[A] with SymmetricTransitionRatio[A])*)(implicit rnd: Random): MixtureProposal[A] with SymmetricTransitionRatio[A] = new MixtureProposal[A](proposals.toIndexedSeq) with SymmetricTransitionRatio[A]
 
   /** mixture of symmetric proposals (mixture distribution) with a transition probability */
-  def fromSymmetricProposalsWithTransition[A](proposals: (Double, ProposalGenerator[A] with TransitionProbability[A] with SymmetricTransition[A])*)(implicit rnd: Random): MixtureProposal[A] with TransitionProbability[A] with SymmetricTransition[A] = new MixtureProposalWithTransition[A](proposals.toIndexedSeq) with SymmetricTransition[A]
+  def fromSymmetricProposalsWithTransition[A](proposals: (Double, ProposalGenerator[A] with TransitionProbability[A] with SymmetricTransitionRatio[A])*)(implicit rnd: Random): MixtureProposal[A] with TransitionProbability[A] with SymmetricTransitionRatio[A] = new MixtureProposalWithTransition[A](proposals.toIndexedSeq) with SymmetricTransitionRatio[A]
```
sschoenborn: this would now become `ProposalGenerator[A] with SymmetricTransition[A]`?

Ghazi-Bouabene: indeed .. commit coming right up :)

PR #155, opened 2016-08-19
Introduce new random object that acts as a default source of randomness

marcelluethi:
A thin wrapper around different random number generators is introduced, which should become the new default source of randomness for all methods in scalismo that need random numbers.

The idea is that every method takes an implicit instance of the new Random class as an argument.

The new mechanism has the following behaviour:
* If the user does not specify an own implicit instance, the default source in the companion object is used.
* If the result should be reproducible, the user can define his/her own implicit instance with a fixed seed.


sschoenborn:
merged manually due to conflicts

PR #158, opened 2016-08-30
meshToDistanceImage uses closest point on surface instead of closest …

Ghazi-Bouabene:
…vertex


Ghazi-Bouabene:
It seems that the SurfaceSpatialIndex.findClosestPoint method is not thread safe. Since this will be used in a parallel collection when evaluating image metrics, this PR should be dropped until that is fixed.

PR #161, opened 2016-09-14
Thread save SpatialIndex with tests. fixes #159

### LINKED ISSUE #159: SurfaceSpatialIndex.findClosestPoint is not thread safe
This function would be very useful to use when computing a distance map from a mesh as it would give a more precise value than the current implementation relying on the closest vertex.

However, distance maps are usually iterated over in parallel when performing registration, hence the issue.

Andreas-Forster:
Introduced ThreadLocal variabales to make the SurfaceSpatialIndex and the DiscreteSpatialIndex thread safe. This pull request fixes #159 without changing the interface. A test is provided that a map and a par.map of sequential requests to a spatial index returns the same results.

In terms of computation speed it seems to make little to no difference when we use ThreadLocal variables for the single threaded case.


Andreas-Forster:
Lost changes of scalismo.Random. Therefore the pull request will be reopened with corrected random number generator.

PR #165, opened 2016-10-19
MeshOperations

Andreas-Forster:
As discussed offline i started this branch as a discussion basis. Yet I only moved the intersect and closestPointOnSurface operations to the MeshOperation class. The important changes are in:
- MeshOperations 
- TriangleMesh

Functionality in MeshOperations:
- Intersect
- ClosestPoint
- MeshBoundaryPredicate

### DIFF
--- a/src/main/scala/scalismo/mesh/MeshOperations.scala
+++ b/src/main/scala/scalismo/mesh/MeshOperations.scala
@@ -15,15 +15,21 @@
  */
 package scalismo.mesh
 
-import scalismo.common.PointId
+import scalismo.common.{ PointId, RealSpace }
 import scalismo.geometry.{ Point, Vector, _3D }
+import scalismo.image.{ DifferentiableScalarImage, ScalarImage }
 import scalismo.mesh.boundingSpheres._
 
 object MeshOperations {
   def apply(mesh: TriangleMesh3D) = new TriangleMesh3DOperations(mesh)
 }
 
-class TriangleMesh3DOperations(mesh: TriangleMesh3D) {
+class TriangleMesh3DOperations(private val mesh: TriangleMesh3D) {
+
+  /**
+   * Calculated data from mesh
+   */
+  private lazy val meshPoints = mesh.pointSet.points.toIndexedSeq
 
   /**
    * Bounding spheres based mesh operations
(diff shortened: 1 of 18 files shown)


src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -0,0 +1,52 @@
...
+class TriangleMesh3DOperations(mesh: TriangleMesh3D) {
+
+  /**
+   * Bounding spheres based mesh operations
+   */
+  private lazy val triangles = BoundingSpheres.triangleListFromTriangleMesh3D(mesh)
+  private lazy val boundingSpheres = BoundingSpheres.createForTriangles(triangles)
+
+  private lazy val intersect: TriangulatedSurfaceIntersectionIndex[_3D] = new LineTriangleMesh3DIntersectionIndex(boundingSpheres, mesh, triangles)
+  def hasIntersection(point: Point[_3D], direction: Vector[_3D]) = intersect.hasIntersection(point, direction)
```
marcelluethi: public methods should state the return type explicitly

src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -0,0 +1,52 @@
...
+  def getIntersectionPoints(point: Point[_3D], direction: Vector[_3D]): Seq[Point[_3D]] = intersect.getIntersectionPoints(point, direction)
+  def getIntersectionPointsOnSurface(point: Point[_3D], direction: Vector[_3D]): Seq[(TriangleId, BarycentricCoordinates)] = intersect.getSurfaceIntersectionPoints(point, direction)
+
+  private lazy val closestPointOnSurface: SurfaceSpatialIndex[_3D] = new TriangleMesh3DSpatialIndex(boundingSpheres, mesh, triangles)
+  def shortestDistanceToSurface(point: Point[_3D]): Double = closestPointOnSurface.getSquaredShortestDistance(point: Point[_3D])
+  def closestPointOnSurfaceWithSquaredDistance(point: Point[_3D]): (Point[_3D], Double) = closestPointOnSurface.getClosestPoint(point)
+  def closestPointOnSurface(point: Point[_3D]): ClosestPoint = closestPointOnSurface.getClosestPointMeta(point)
```
marcelluethi: here the closest point type, which seems to be used only internally leaks into the public interface. Is this done intentionally?

src/main/scala/scalismo/mesh/TriangleMesh.scala
```diff
@@ -85,6 +85,8 @@ case class TriangleMesh3D(pointSet: UnstructuredPointsDomain[_3D], triangulation
   val triangles = triangulation.triangles
   val cells = triangles
 
+  lazy val operations = MeshOperations(this)
```
marcelluethi: It would be good to explicitly state here that the return type is a TriangleMeshOperation (and not a generic MeshOperation)

src/main/scala/scalismo/mesh/boundingSpheres/SurfaceIntersectionIndex.scala
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * Creates SurfaceDistance for a TriangleMesh3D.
+   */
+  def fromTriangleMesh3D(mesh: TriangleMesh3D): TriangulatedSurfaceIntersectionIndex[_3D] = {
+
+    // build triangle list (use only Vector[_3D], no Points)
+    val triangles = mesh.triangulation.triangles.map { t =>
+
+      val a = mesh.pointSet.point(t.ptId1).toVector
```
marcelluethi: is it necessary to convert to a vector here? You loose important type information

src/main/scala/scalismo/mesh/MeshOperations.scala
Andreas-Forster: This is made intentionally. You can pattern match the type to find out if it is a vertex, a point on a line or a point in a triangle. This will be useful in some applications.
Maybe a point on a surface as (TriangleId, BarycentricCoordinates) might be worth to be considered as an additional method.

src/main/scala/scalismo/mesh/boundingSpheres/SurfaceIntersectionIndex.scala
Andreas-Forster: it makes the internal computation much more readable as many toVector and toPoint can be omitted. Beside the readability it reduces the computational burden, if the JVM does not remove these conversations.

src/main/scala/scalismo/mesh/TriangleMesh.scala
Andreas-Forster: Fixed.

src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -0,0 +1,52 @@
...
+
+class TriangleMesh3DOperations(mesh: TriangleMesh3D) {
+
+  /**
+   * Bounding spheres based mesh operations
+   */
+  private lazy val triangles = BoundingSpheres.triangleListFromTriangleMesh3D(mesh)
+  private lazy val boundingSpheres = BoundingSpheres.createForTriangles(triangles)
+
+  private lazy val intersect: TriangulatedSurfaceIntersectionIndex[_3D] = new LineTriangleMesh3DIntersectionIndex(boundingSpheres, mesh, triangles)
```
sschoenborn: `LineTriangleMesh3DIntersectionIndex` is a case class, no `new` needed ... but I would suggest to change it into an ordinary class, see comments there

src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -0,0 +1,52 @@
...
+  def hasIntersection(point: Point[_3D], direction: Vector[_3D]) = intersect.hasIntersection(point, direction)
+  def getIntersectionPoints(point: Point[_3D], direction: Vector[_3D]): Seq[Point[_3D]] = intersect.getIntersectionPoints(point, direction)
+  def getIntersectionPointsOnSurface(point: Point[_3D], direction: Vector[_3D]): Seq[(TriangleId, BarycentricCoordinates)] = intersect.getSurfaceIntersectionPoints(point, direction)
+
+  private lazy val closestPointOnSurface: SurfaceSpatialIndex[_3D] = new TriangleMesh3DSpatialIndex(boundingSpheres, mesh, triangles)
```
sschoenborn: `new` again - here, I also suggest to turn `TriangleMesh3DSpatialIndex` into an ordinary class

src/main/scala/scalismo/mesh/boundingSpheres/SurfaceIntersectionIndex.scala
```diff
@@ -0,0 +1,140 @@
...
+    val bs = BoundingSpheres.createForTriangles(triangles)
+
+    // new distance object
+    new LineTriangleMesh3DIntersectionIndex(bs, mesh, triangles)
+
+  }
+
+}
+
+private[mesh] case class LineTriangleMesh3DIntersectionIndex(private val boundingSphere: BoundingSphere,
```
sschoenborn: I suggest to remove `case` here: the class does not support the "value"-type equality and hash semantics one would expect due to the `BoundingSphere` member (usage in `TriangleMesh3DOperations` is already with `new` anyway)

src/main/scala/scalismo/mesh/boundingSpheres/SurfaceSpatialIndex.scala
```diff
@@ -111,9 +112,9 @@ private case class ClosestPointMeta(val distance2: Double,
 /**
  * Surface distance implementation for TriangleMesh3D.
  */
-private case class TriangleMesh3DSpatialIndex(bs: BoundingSphere,
+private[mesh] case class TriangleMesh3DSpatialIndex(bs: BoundingSphere,
```
sschoenborn: I suggest to remove case here: the class does not support the "value"-type equality and hash semantics one would expect due to the BoundingSphere member (usage in TriangleMesh3DOperations is already with new anyway)

src/main/scala/scalismo/mesh/boundingSpheres/SurfaceSpatialIndex.scala
```diff
@@ -143,16 +144,16 @@ private case class TriangleMesh3DSpatialIndex(bs: BoundingSphere,
...
 
       case ON_LINE => res.get().idx match {
-        case (0, _) => ClosestPointOnLine(res.get().pt.toPoint, res.get().distance2, (triangle.pointIds(0).id, triangle.pointIds(1).id), res.get().bc.a)
-        case (1, _) => ClosestPointOnLine(res.get().pt.toPoint, res.get().distance2, (triangle.pointIds(1).id, triangle.pointIds(2).id), res.get().bc.a)
-        case (2, _) => ClosestPointOnLine(res.get().pt.toPoint, res.get().distance2, (triangle.pointIds(2).id, triangle.pointIds(0).id), res.get().bc.a)
+        case (0, _) => ClosestPointOnLine(res.get().pt.toPoint, res.get().distance2, (PointId(triangle.pointIds(0).id), PointId(triangle.pointIds(1).id)), res.get().bc.a)
+        case (1, _) => ClosestPointOnLine(res.get().pt.toPoint, res.get().distance2, (PointId(triangle.pointIds(1).id), PointId(triangle.pointIds(2).id)), res.get().bc.a)
+        case (2, _) => ClosestPointOnLine(res.get().pt.toPoint, res.get().distance2, (PointId(triangle.pointIds(2).id), PointId(triangle.pointIds(0).id)), res.get().bc.a)
         case _ => throw (new RuntimeException("not a valid line index"))
```
sschoenborn: unnecessary `(...)` (sorry, but all these things turn bright yellow in my IDEA :))

src/main/scala/scalismo/mesh/boundingSpheres/SurfaceSpatialIndex.scala
```diff
@@ -143,16 +144,16 @@ private case class TriangleMesh3DSpatialIndex(bs: BoundingSphere,
...
       }
 
-      case IN_TRIANGLE => ClosestPointInTriangle(res.get().pt.toPoint, res.get().distance2, lastIdx.get().idx, (res.get().bc.a, res.get().bc.b))
+      case IN_TRIANGLE => ClosestPointInTriangle(res.get().pt.toPoint, res.get().distance2, TriangleId(lastIdx.get().idx), BarycentricCoordinates(1.0 - res.get().bc.a - res.get().bc.b, res.get().bc.a, res.get().bc.b))
 
       case _ => throw (new RuntimeException("not a valid PointType"))
```
sschoenborn: unnecessary `(...)` (sorry, but all these things turn bright yellow in my IDEA :))

Andreas-Forster:
Now mesh operations present in scalismo and an additional library were merged in MeshOperations. All usages of functions from the Mesh object are removed within scalismo. The functions from Mesh are marked deprecated as they have equivalents in MeshOperations now.

src/main/scala/scalismo/mesh/MeshOperations.scala
Ghazi-Bouabene: Shouldn't the closestPointOnSurfaceWithSquaredDistance method also return a (ClosestPoint, Double) in that case?

Andreas-Forster: No. The ClosestPoint is the most informative answer you can get for the query. The other methods are only convenient methods if you are interested only in less information. (Name will be changed for 1st method)
ClosestPoint.distance2  <= shortestDistanceToSurfaceSquared
(ClosestPoint.point, ClosestPoint.distance2) <= closestPointOnSurfaceWithSquaredDistance
ClosestPoint <= closestPointOnSurface

Andreas-Forster: If you look at it from the side of a Triangulated Surface it might be also interesting to know in which triangle it lies and which barycentric coordinates it has or if it is one of the verticies or on a line connecting two vertices. This is solved by the type ClosestPoint. Maybe we can rename this to ClosestPointDescription. Would that help?

src/main/scala/scalismo/mesh/boundingSpheres/ClosestPoint.scala
```diff
@@ -28,14 +30,14 @@ class ClosestPoint(val point: Point[_3D],
 
 case class ClosestPointIsPoint(override val point: Point[_3D],
```
Andreas-Forster: Done

src/main/scala/scalismo/mesh/MeshOperations.scala
Andreas-Forster: Renaming done.

PR #168, opened 2016-11-16
Use sbt-git for versioning

marcelluethi:
Thanks to the use of sbt-git, the build version can be determined based on the git version. More precisely, git describe is used to build a unique version, based on the last tagged release and a short hash of the current commit.


sschoenborn:
I like this very much! It solves the long standing problem of non-reproducible development dependencies to use recent features while still keeping releases at a meaningful pace.

PR #169, opened 2016-11-21
Removed scalariform.sbt

Andreas-Forster:
The reason to remove this file is that this produces published versions of scalismo that are not contained in the git-history of commits. As we now want to use single commit based artifacts, this is no longer an option.

### DIFF
--- a/scalariform.sbt
+++ b/scalariform.sbt
@@ -1,1 +1,0 @@
-scalariformSettings


clangguth:
I disagree.

The purpose of this is to reformat *before* committing, to ensure a consistent style for all commits.
If everybody did that, we would see way less noise in the commit diffs (mostly whitespace changes), cluttering and effectively hiding where actual logic was changed.

If scalariform were used the way it's meant to ("sbt compile; git commit" from the command-line), we wouldn't have that problem at all. But it requires the discipline to do so.

If you remove this file, I predict that the issue will get even worse, because then everybody will be pushing things with their own formatting, and the noise would just increase.

Andreas-Forster:
I fully agreee that PR (#170) is the better solution. I recommend that every developer should locally install in the repository git-hooks the following two line:

```
sbt scalariformFormat 
sbt test:scalariformFormat
```

PR #170, opened 2016-11-21
change travis build such that it fails if source is not properly formatted

marcelluethi:
This change causes Travis to fail if the source is not properly formatted (according to scalariform).  This will make it much easier to spot when the formatting for a  proposed change is not correct. 

In the case of an incorrect formatting, a PR should not be merged. This should help to avoid situations, as the one described in PR #169, but also more generally lead to more clear commits/diffs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ env:
  - SCALISMO_PLATFORM=linux64
 
 ## Caching seems to cause problems of its own, so we leave it commented for now.
-## We use travis_wait instead (see below in the script: section) to give the build more time. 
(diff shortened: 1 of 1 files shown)


clangguth:
Haha, this is an awesome idea! Thumbs up!

I can't judge if it'll work as intended, but it seems to make sense to me.

sschoenborn:
I like this. I think scalariform only makes sense if it is somehow enforced, e.g. "soft" through failed Travis builds.

Should the build fail faster, before running all the tests?

sschoenborn:
To still see what tests are ok, even with bad formatting, we let Travis run all tasks

PR #177, opened 2017-01-17
Improved accuracy when computing nystrom approximation. Fixes #176.

### LINKED ISSUE #176: Inaccurate eigenfunction computation in Nyström approximation
As an intermediate result in the Nyström method, an Eigendecomposition of the kernel matrix, defined on a subset of the domain points has to be computed. For efficiency reasons, this is currently done using the Pivoted Cholesky Decomposition. In the current implementation, the accuracy is not specified as a relative tolerance, but only a fixed number of basis vectors are computed. This can lead to bad approximations of the true eigenvectors. 

A fix is to use a relative tolerance in the computation.

marcelluethi:
There are now two variants of the Nystrom method. In the first variant the user does specify
the number of leading eigenfunctions that should be computed. In this case a RandomSVD is used for
computing the leading eigenvectors of the Kernel matrix. If the user does not specify the number of
eigenfunctions, the method chooses the number of eigenfunctions such that a certain relative error tolerance
is fulfilled.

### DIFF
--- a/src/main/scala/scalismo/kernels/Kernel.scala
+++ b/src/main/scala/scalismo/kernels/Kernel.scala
@@ -18,7 +18,7 @@ package scalismo.kernels
 import breeze.linalg.{ DenseMatrix, DenseVector, diag, pinv }
 import scalismo.common._
 import scalismo.geometry._
-import scalismo.numerics.PivotedCholesky.NumberOfEigenfunctions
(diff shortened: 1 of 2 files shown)


marcelluethi:
Do not merge this PR yet.
It might be a better approach to always approximate the eigenvectors to a certain accuracy using the PivotedCholesky decomposition. The computation time is usually dwarfed by the actual interpolation.

marcelluethi:
superseded by #179

PR #178, opened 2017-01-18
Zero mean gaussianprocess

marcelluethi:
A very common assumption when defining Gaussian process models is that the Gaussian process is zero mean. This is currently quite cumbersome to specify, and the code becomes unreadable. 
This PR proposes to add an additional factory function to the ```GaussianProcess``` object, which makes it possible to create a zero-mean Gaussian process  by specifying only the covariance function

Example: Instead of 
```
GaussianProcess[_3D, Vector[_3D]](Field(domain, _ => Vector(0.0, 0.0, 0.0)), kernel)
```
it is now possible to define 
```
GaussianProcess[_3D, Vector[_3D]](kernel)
```


Andreas-Forster:
Thx for the contribution. This convenience method will be used quite often.

PR #183, opened 2017-01-24
Harmonized and improved interface for landmark registration

marcelluethi:
* All methods can now be called using the same arguments.
* The rotation center needs to be specified explicitly in all the methods.
* IndexedSeq are replaced with Seq
* SimilarityLandmarkRegistration methods now return proper similarity transform

### DIFF
--- a/src/main/scala/scalismo/registration/LandmarkRegistration.scala
+++ b/src/main/scala/scalismo/registration/LandmarkRegistration.scala
@@ -76,6 +76,11 @@ object LandmarkRegistration {
     LandmarkRegistration.rigid3DLandmarkRegistration(landmarksPairs.toIndexedSeq, center)
   }
 
+  @deprecated("Calling rigid3DLandmarkRegistration without a center is deprecated. Please specify the center", "0.15")
+  def rigid3DLandmarkRegistration(originalLms: Seq[Landmark[_3D]], targetLms: Seq[Landmark[_3D]]): RigidTransformation[_3D] = {
+    LandmarkRegistration.rigid3DLandmarkRegistration(originalLms, targetLms, center = origin3D)
+  }
+
   /**
    * Returns a rigid transformation mapping the original landmarks  (first elements of the tuples) into the corresponding target points (second elements of the tuples).
    *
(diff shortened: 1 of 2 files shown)


src/main/scala/scalismo/registration/LandmarkRegistration.scala
```diff
@@ -97,16 +97,93 @@ object LandmarkRegistration {
    *
    * @param landmarks sequence of corresponding landmarks
    */
+  def rigid3DLandmarkRegistration(landmarks: Seq[(Point[_3D], Point[_3D])]): RigidTransformation[_3D] = rigid3DLandmarkRegistration(landmarks, origin3D)
```
Ghazi-Bouabene: Center is defaulted here to origin. Does not match with the strategy to explicitly state the center everywhere.

src/main/scala/scalismo/registration/LandmarkRegistration.scala
```diff
@@ -97,16 +97,93 @@ object LandmarkRegistration {
...
+    rigidSpace.transformForParameters(optimalParameters)
+  }
+
+  /**
+   * Returns a rigid transformation mapping the original landmarks (first elements of the tuples) into the corresponding target points (second elements of the tuples).
+   *
+   * @param landmarks sequence of corresponding landmarks
+   * @param center - center of the rotation
+   */
+  def rigid2DLandmarkRegistration(originalLms: Seq[Landmark[_2D]], targetLms: Seq[Landmark[_2D]], center: Point[_2D]): RigidTransformation[_2D] = {
```
Ghazi-Bouabene: As the matching or landmarks is done by name here and not by the ordering, we could relax the type constraint from Seq to perhaps Traversable, as the "sequentiality" plays no role here.

src/main/scala/scalismo/registration/LandmarkRegistration.scala
```diff
@@ -97,16 +97,93 @@ object LandmarkRegistration {
...
+
+  /**
+   * Returns a similarity transformation mapping the original landmarks (first elements of the tuples)
+   * into the corresponding target points (second elements of the tuples).
+   *
+   * Attention: the center of rotation used for the rigid transformation here is the origin of the 3D space
+   *
+   * @param landmarks sequence of corresponding landmarks
+   */
+  def similarity3DLandmarkRegistration(landmarks: Seq[(Point[_3D], Point[_3D])]): ParametricTransformation[_3D] = similarity3DLandmarkRegistration(landmarks, origin3D)
```
Ghazi-Bouabene: Same comment as above, the center or rotation is defaulted here which does not match with the strategy to make it mandatory.

src/main/scala/scalismo/registration/LandmarkRegistration.scala
```diff
@@ -97,16 +97,93 @@ object LandmarkRegistration {
...
+    SimilarityTransformation(scalingTransform, rigidTransform)
+  }
+
+  /**
+   * Returns a similarity transformation mapping the original landmarks (first elements of the tuples) into the corresponding target points (second elements of the tuples).
+   *
+   * @param landmarks sequence of corresponding landmarks
+   * @param center - center of the rotation
+   */
+  def similarity3DLandmarkRegistration(originalLms: Seq[Landmark[_3D]], targetLms: Seq[Landmark[_3D]], center: Point[_3D]): SimilarityTransformation[_3D] = {
```
Ghazi-Bouabene: Same comment as above, maybe use Traversable instead of Seq

PR #184, opened 2017-02-14
Introducing a convenience method to reduce the rank of Gaussian process models

marcelluethi:
A common task when working with shape models is to reduce the rank of a model (i.e. to create a new model with only the leading basis functions).

While this was always possible in scalismo, it was so well hidden and cumbersome that the uninitiated would never find out how to do it. This PR proposes to add a method ```truncate``` to the ```StatisticalMeshModel``` and the ```(Discrete)LowrankGaussianProcess``` classes. 

To obtain a new statistical model, which is only represented using the first 5 basis function, the following code can be used:
```scala
val reducedRankModel = model.truncate(newRank = 5)
```

### DIFF
--- a/src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
+++ b/src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
@@ -119,6 +119,15 @@ case class DiscreteLowRankGaussianProcess[D <: Dim: NDSpace, Value] private[scal
   }
 
   /**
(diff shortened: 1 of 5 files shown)


sschoenborn:
Looks very convenient. We actually did reimplement this in the past because we were not aware of the functionality.

Test failed:
```
[info] - can reproduce the covariance function from random samples *** FAILED ***
[info]   0.7121637152263456 was not 1.0000228887834257 plus or minus 0.2 (KernelTests.scala:95) 
```

This looks unrelated to the pull request, I think

PR #186, opened 2017-02-22
Changed LandmarkIO interface to use Seq and removed deprecated methods.

marcelluethi:
The LandmarkIO methods were among the few methods that require an immutable seq. While it is technically correct to use an immutable seq, it is inconsistent with the rest of scalismo and causes difficult to read compiler errors. This PR thus proposes to change these into "normal" seqs.

At the same time, the deprecated methods (which were there since v0.10) are removed.

### DIFF
--- a/src/main/scala/scalismo/io/LandmarkIO.scala
+++ b/src/main/scala/scalismo/io/LandmarkIO.scala
@@ -92,7 +92,7 @@ object LandmarkIO {
 
   def writeLandmarksJson[D <: Dim: NDSpace](landmarks: Seq[Landmark[D]], file: File): Try[Unit] = writeLandmarksJsonToStream[D](landmarks, new FileOutputStream(file))
 
-  private def writeLandmarksJsonToStreamP[D <: Dim: NDSpace](stream: OutputStream, landmarks: Seq[Landmark[D]])(implicit e: JsonFormat[Landmark[D]]): Try[Unit] = {
+  private def writeLandmarksJsonToStreamP[D <: Dim: NDSpace](landmarks: Seq[Landmark[D]], stream: OutputStream)(implicit e: JsonFormat[Landmark[D]]): Try[Unit] = {
     val writer = new PrintWriter(stream, true)
     val result = Try {
       writer.println(landmarks.toJson.toString())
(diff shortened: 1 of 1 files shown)


src/main/scala/scalismo/io/LandmarkIO.scala
```diff
@@ -91,15 +90,9 @@ object LandmarkIO {
...
-  def writeLandmarksJson[D <: Dim: NDSpace](file: File, landmarks: immutable.Seq[Landmark[D]]): Try[Unit] = writeLandmarksJson(landmarks, file)
+  def writeLandmarksJson[D <: Dim: NDSpace](landmarks: Seq[Landmark[D]], file: File): Try[Unit] = writeLandmarksJsonToStream[D](landmarks, new FileOutputStream(file))
 
-  @deprecated("Use method with the same name and inverted order of parameters. This method wil be removed in future versions.", "0.10.0")
-  def writeLandmarksJsonToStream[D <: Dim: NDSpace](stream: OutputStream, landmarks: immutable.Seq[Landmark[D]]): Try[Unit] = writeLandmarksJsonToStream(landmarks, stream)
-
-  def writeLandmarksJson[D <: Dim: NDSpace](landmarks: immutable.Seq[Landmark[D]], file: File): Try[Unit] = writeLandmarksJsonToStream[D](landmarks, new FileOutputStream(file))
-
-  private def writeLandmarksJsonToStreamP[D <: Dim: NDSpace](stream: OutputStream, landmarks: immutable.Seq[Landmark[D]])(implicit e: JsonFormat[Landmark[D]]): Try[Unit] = {
+  private def writeLandmarksJsonToStreamP[D <: Dim: NDSpace](stream: OutputStream, landmarks: Seq[Landmark[D]])(implicit e: JsonFormat[Landmark[D]]): Try[Unit] = {
```
Andreas-Forster: Shouldn't the arguments be swaped here to be consistent with the other write* methods?

Ghazi-Bouabene:
I also agree.

src/main/scala/scalismo/io/LandmarkIO.scala
marcelluethi: True. Should be fixed.

PR #187, opened 2017-03-13
Clean Random & make RandBasis available

sschoenborn:
The `Random` object is meant to unify how randomness is used and ensures a seeded state. However, it was not possible to get a breeze `RandBasis` due to it being private. When working mit more than `breezeUniform` and `breezeGaussian` this is required, e.g. in 

```scala
import breeze.stats.distributions._

val rnd = Random(9)
import rnd.breezeRandBasis
val gauss = Exponential(2.0)
gauss.draw() // should yield: 0.48878484018783064
```

- Also fixed a typo in `breezeUniform`
- Ranamed parameter `sigma` of `breezeGaussian` to avoid confusion - it is a standard deviation, not a variance
- Use `System.nanoTime()` for more fine-grained seeding

### DIFF
--- a/src/main/scala/scalismo/numerics/RandomSVD.scala
+++ b/src/main/scala/scalismo/numerics/RandomSVD.scala
@@ -17,8 +17,9 @@ package scalismo.numerics
 
 import breeze.linalg.qr.QR
 import breeze.linalg.svd.SVD
-import breeze.linalg.{ diag, DenseMatrix, DenseVector, norm }
-import scalismo.utils.{ Random, Benchmark }
+import breeze.linalg.{ DenseMatrix, DenseVector, diag, norm }
+import breeze.stats.distributions.Gaussian
+import scalismo.utils.{ Benchmark, Random }
 
 /**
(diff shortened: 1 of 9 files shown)


marcelluethi:
@sschoenborn 
Thanks for this fix. I agree with most changes, but find that the new interface is inconsistent. Either we should access the object via the randBasis, and then deprecate the ```breezeUniform``` method (my preference), or we add for all interesting distributions accessor methods.  

What do you think?

sschoenborn:
I would then prefer the first solution: deprecate the direct methods `breezeUniform` / `breezeGaussian` and promote the direct use of `breezeRandBasis`

sschoenborn:
by the way: breeze 0.13 has a direct `RandBasis.withSeed(seed)` (and fixes a wrong pdf in `MultivariateGaussian`)

marcelluethi:
@sschoenborn Awesome. Thanks a lot for the changes. I will try to upgrade the dependency to breeze 0.13. This would also give us compatibility with scala 2.12.

PR #188, opened 2017-03-13
Use breeze v0.13

marcelluethi:
Update breeze dependency to v0.13.


sschoenborn:
Great! with breeze 0.13 we can start towards scala 2.12

PR #189, opened 2017-03-14
depend on scalismo-native 4.0.0 (on jcenter)

marcelluethi:
Update to latest version of scalismo-native, which is now hosted on JCentral.


sschoenborn:
works nicely - with a much leaner resolver structure

PR #190, opened 2017-03-14
Update to scala 2.12.1

sschoenborn:
Major scala update to version 2.12.1. The new version with many cool new [features in 2.12.0](http://www.scala-lang.org/news/2.12.0) (and less impressive [2.12.1](http://www.scala-lang.org/news/2.12.1)). This will make scalismo leaner and slightly easier to use.

Be aware: you need java 8 now!

Only few adaptation were needed:

- Some dependency updates (thanks Marcel)
- explicit types in `MixtureProposal`
- Travis update to use scala 2.12.1 (already on jdk8)

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -13,14 +13,14 @@ import com.typesafe.sbt.SbtGit.useJGit
 
 object BuildSettings {
   val buildOrganization = "ch.unibas.cs.gravis"
-  val buildScalaVersion = "2.11.7"
+  val buildScalaVersion = "2.12.1"
 
   val publishURL = Resolver.file("file", new File("/export/contrib/statismo/repo/public"))
 
   val buildSettings = Defaults.defaultSettings ++ Seq(
     organization := buildOrganization,
     scalaVersion := buildScalaVersion,
-    crossScalaVersions := Seq("2.10.5", "2.11.7"),
+    crossScalaVersions := Seq("2.11.7"),
     javacOptions ++= Seq("-source", "1.6", "-target", "1.6"),
     scalacOptions ++= Seq("-encoding", "UTF-8", "-Xlint", "-deprecation", "-unchecked", "-feature", "-target:jvm-1.6")
   )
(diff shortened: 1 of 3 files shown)


project/Build.scala
```diff
@@ -13,14 +13,14 @@ import com.typesafe.sbt.SbtGit.useJGit
 
 object BuildSettings {
   val buildOrganization = "ch.unibas.cs.gravis"
-  val buildScalaVersion = "2.11.7"
+  val buildScalaVersion = "2.12.1"
```
marcelluethi: I think we should list both versions (2.11.7 and 2.12.1) as ```crossScalaVersions``` below and use 2.11.7 as the default version for the compilation. In this way we are not likely to use language features that are not supported by scala 2.11.

sschoenborn:
I have put 2.11.8 as main version (2.11.7 is automatically upgraded to 2.11.8 by sbt by a dependency anyway) and both as cross versions

I also updated target settings to 1.8 rather than 1.6 - not quite sure whether this is relevant and ok?

marcelluethi:
@sschoenborn I just added a commit, which selects the target jvm based on the scalac version. This should make it possible to continue to use scalismo with java < 1.8.  If this seems reasonable to you, we can merge this PR.

sschoenborn:
Looks good!

Am 14.03.2017 um 21:17 schrieb Marcel Luethi <notifications@github.com<mailto:notifications@github.com>>:


@sschoenborn<https://github.com/sschoenborn> I just added a commit, which selects the target jvm based on the scalac version. This should make it possible to continue to use scalismo with java < 1.8. If this seems reasonable to you, we can merge this PR.

clangguth:
That's pretty awesome :-)

While you're at it, some minor stuff: please remove the year in the copyright notices. In Build.scala it says 2015, and in the headers I think it's 2016.

Anyway, the year is not needed for a Copyright claim to be valid.

marcelluethi:
@clangguth The headers in the files are auto-generated (it is done by an sbt plugin). I will look into it and if necessary regenerate the headers for all files in the project in a separate PR.

PR #191, opened 2017-03-15
Cleaner logging for sampling

sschoenborn:
Logging in `sampling` was a bit messy: Chains allowed to run multiple iterators (multiple runs with the same settings but no communication), however, logging was on object level only. So multiple runs inferred in the same logger.

I cleaned this and made the loggers attach to individual runs rather than the complete object. You can now start a "logged" iterator in a `Metropolis`/`MetropolisHastings` algorithm. Also a `MarkovChain` has no global logger any more but can attach one to the chain iterator.


Andreas-Forster:
I like this PR as moving the logging to the iterator places it at a better location.

Issue #193, opened 2017-03-27
Memory problems with long running jobs that use vtk


marcelluethi:
Long running jobs that do a lot of MeshIO run out of memory at some point. 
The problem is most likely that memory is leaked when using VTK.

marcelluethi:
Fixed in #194

PR #194, opened 2017-03-27
start vtk garbage collector on initialize.

marcelluethi:
This PR addresses the problem that long running jobs that use ```MeshIO``` run out of memory at some point (issue #193). 

The fix is to start the VTK Garbage Collector explicitly on calling ```scalismo.initialize()```.

### DIFF
--- a/src/main/scala/scalismo/package.scala
+++ b/src/main/scala/scalismo/package.scala
@@ -13,7 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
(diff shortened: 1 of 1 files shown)


clangguth:
I'm afraid it's not that easy, because you may also have to ensure that the VTK garbage collector is stopped, so that your program will reliably exit.

See the scalismo.ui.rendering.Rendering class for how this is currently implemented in the UI. That code would also have to be adjusted accordingly (to take into account the new behavior of initialize() ).

PR #195, opened 2017-03-29
Explicit tolerance for merging in STL

marcelluethi:
It seems to have happened in some cases, that the STL reader merged points, which were very close to each other (but not identical). As this destroy any point-to-point correspondence that might exist among meshes, such a merge should never happen. To prevent this, we set the tolerance of the stl reader to zero.

### DIFF
--- a/src/main/scala/scalismo/io/MeshIO.scala
+++ b/src/main/scala/scalismo/io/MeshIO.scala
@@ -250,6 +250,16 @@ object MeshIO {
   private def readSTL(file: File, correctMesh: Boolean = false): Try[TriangleMesh[_3D]] = {
     val stlReader = new vtkSTLReader()
     stlReader.SetFileName(file.getAbsolutePath)
+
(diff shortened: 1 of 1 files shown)


src/main/scala/scalismo/io/MeshIO.scala
```diff
@@ -250,6 +250,16 @@ object MeshIO {
   private def readSTL(file: File, correctMesh: Boolean = false): Try[TriangleMesh[_3D]] = {
     val stlReader = new vtkSTLReader()
     stlReader.SetFileName(file.getAbsolutePath)
+
+    stlReader.MergingOn()
```
Ghazi-Bouabene: Why not use MergingOff()? Would we still want triangle merging?

marcelluethi: If we choose MergingOff, then no points are merged, even not the identical ones which are only there because the same point is defined in two different triangles.

PR #197, opened 2017-04-12
Corrected computation of  Procrustes distance (fixes #196)

### LINKED ISSUE #196: procrustesDistance uses the closest point instead of corresponding point
```MeshMetrics.procrustesDistance``` does not compute the Procrustes distance correctly. It correctly computes the rigid alignment that minimizes the squared error between corresponding points. However, it then uses the method ```avgDistance``` to compute the distances, which computes the distance to the closest point and not the corresponding point.

### PR DESCRIPTION
**Pull request by marcelluethi**

### CODE REVIEW

File: src/main/scala/scalismo/mesh/MeshMetrics.scala [hunk 1]
```diff
@@ -42,16 +42,23 @@ object MeshMetrics {
...
     val landmarks = m1.pointSet.points.toIndexedSeq zip m2.pointSet.points.toIndexedSeq
     val t = LandmarkRegistration.rigid3DLandmarkRegistration(landmarks, Point(0, 0, 0))
-    val m1w = m1.transform(t)
-    avgDistance(m1w, m2)
+    val m1Aligned = m1.transform(t)
+    var accumulatedDistance = 0.0
+    for ((pt1, pt2) <- m1Aligned.pointSet.points.zip(m2.pointSet.points)) {
+      accumulatedDistance += (pt1 - pt2).norm2
+    }
+    Math.sqrt(accumulatedDistance / m1.pointSet.numberOfPoints)
```
Ghazi-Bouabene: Shouldn't this be the mean of all square roots (or of (pt1 - pt2).norm), instead of the square root of the mean squared distance?

marcelluethi: @Ghazi-Bouabene The use of the squared norm is correct (see the [definition](https://en.wikipedia.org/wiki/Procrustes_analysis) here). However, in the original formula the differences are not divided by the number of points (I guess because in classical shape analysis, the number of landmarks is anyway always fixed). Putting it inside means we are essentially computing the root mean squared distance, which seems natural to me. Would there be an advantage (such as easier interpretation) to have it outside of the root?

Ghazi-Bouabene: According to the same definition, there would be uniform scaling of the shape involved as well. Combined with the normalization by the number of points inside the square root, we would be far from the standard definition of the Procrustes distance. Therefore we could either rename it, or simply omit it.

marcelluethi:
I changed the base branch of this PR to develop. As the solution will most likely change the interface, it cannot be fixed for release-0.15.

marcelluethi:
I will close this Pull Request without merging it. We will need to take another attempt to fix the problem with the Procrustes distance

PR #198, opened 2017-04-13
use closestPointOnSurface for computing binary image from mesh

marcelluethi:
So far, the computation of a binary image involved checking for every image point the angle to the normal at the closest vertex point. This PR proposes to replace this with the closest point on the surface. Especially for meshes with a low resolution, this should increase the accuracy.

### DIFF
--- a/src/main/scala/scalismo/mesh/MeshOperations.scala
+++ b/src/main/scala/scalismo/mesh/MeshOperations.scala
@@ -99,12 +99,28 @@ class TriangleMesh3DOperations(private val mesh: TriangleMesh3D) {
    *
    */
   def toBinaryImage: ScalarImage[_3D] = {
+
+    val meshOps = mesh.operations
+
     def inside(pt: Point[_3D]): Short = {
-      val closestMeshPt = mesh.pointSet.findClosestPoint(pt)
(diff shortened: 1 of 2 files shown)


src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -99,12 +99,28 @@ class TriangleMesh3DOperations(private val mesh: TriangleMesh3D) {
...
+        case ClosestPointInTriangle(closestPoint, dist, triangleId, bcc) => {
+          (closestPoint, mesh.vertexNormals.onSurface(triangleId, bcc))
+        }
+        case ClosestPointOnLine(closestPoint, _, (id1, id2), bc) => {
+          val normalPt1 = mesh.vertexNormals(id1)
+          val normalPt2 = mesh.vertexNormals(id2)
+          val averagedNormal = (normalPt1 * bc) + (normalPt2 * (1.0 - bc))
+          (closestPoint, averagedNormal / averagedNormal.norm)
+        }
+        case _ => {
```
Andreas-Forster: The wildcard should be replaced by ClosestPointIsVertex.

marcelluethi: @Andreas-Forster As this is only a stylistic issue, which will anyway be addressed when we fix issue #201, I will merge the PR.

PR #199, opened 2017-04-19
Fix SilentLogger's apply method

Andreas-Forster:
In the class `SilentLogger` the `apply` method is written without parentheses. This leads to the following behaviour:
```
val type: SilentLogger.type = SilentLogger
val logger: SilentLogger[A] = SilentLogger.apply
// val logger2: SilentLogger[A] = SilentLogger() → does not compile
```
The last line is working when we add parentheses which is the content of the pull request.


marcelluethi:
@Andreas-Forster should this fix go into 0.15 or are you okay to have it on develop only?

Andreas-Forster:
@marcelluethi Would be great if this is possible.

sschoenborn:
Having it in 0.15 would be great for the tutorial

PR #200, opened 2017-04-19
More general interface for Metric

marcelluethi:
The method ```takeDerivativeWRTToTransform``` returned a function ```Point[_2D]=>Option[DenseVector[Double]]```, which can be used to compute the derivative of the metric at every point.  This exposes too much detail of the internal working of the metric to the outside (e.g. that the final gradient value is obtained by integrating over all the points of the domain). This is not the case for all metrics (i.e. the MutualInformation metric cannot be computed by integrating over
points).

We therefore changed the interface and gradient computation mechanism, such that the metric now computes the full gradient (including the derivative of the transformation). The consequence is, that the metric now needs to know about the TransformationSpace, in order to compute this full gradient.


src/main/scala/scalismo/registration/ImageMetric.scala
```diff
@@ -0,0 +1,120 @@
...
+    transformationSpace: TransformationSpace[D], params: DenseVector[Double]): DenseVector[Double] = {
+    computeDerivative(fixedImage, movingImage, transformationSpace, params, Integrator(sampler))
+  }
+
+  override def computeValueAndDerivative(fixedImage: ScalarImage[D], movingImage: DifferentiableScalarImage[D],
+    transformationSpace: TransformationSpace[D], params: DenseVector[Double]): ValueAndDerivative = {
+
+    // We create a new sampler, which always returns the same points. In this way we can make sure that the
+    // same sample points are used for computing the value and the derivative
+    val sampleOnceSampler = new Sampler[D] {
```
Ghazi-Bouabene: Wouldn't this be called for every new parameter vector evaluation? It could be moved out to the constructor.

src/main/scala/scalismo/registration/ImageMetric.scala
```diff
@@ -0,0 +1,120 @@
...
+    val movingImageGradient = movingImage.differentiate
+    val warpedImage = movingImage.compose(transform) // compute the derivative of the cost function
+    val dDMovingImage = (fixedImage - warpedImage) * (-2.0 / sampler.volumeOfSampleRegion)
+
+    val dTransformSpaceDAlpha = transformationSpace.takeDerivativeWRTParameters(params)
+
+    val fullMetricGradient = (x: Point[D]) => {
+      val domain = Domain.intersection(fixedImage.domain, dDMovingImage.domain)
+      if (domain.isDefinedAt(x))
+        Some(dTransformSpaceDAlpha(x).t * (movingImageGradient(transform(x)) * dDMovingImage(x).toDouble).toBreezeVector)
```
Ghazi-Bouabene: One advantage of the previous interface was that an eventual metric developer would only have to worry about differentiating the expression of the metric itself, the rest of the chain rule being taken care of by the Registration code. In this particular case, one would have to pay attention to to include the dTransformSpaceDAlpha.

Ghazi-Bouabene:
I agree with the changes, with the only remark being that the SampleOnceSampler creation should be moved out to the constructor of the MeanSquaresMetric

src/main/scala/scalismo/registration/ImageMetric.scala
marcelluethi: @Ghazi-Bouabene  This is intentional. The idea here is that the same sample is used for computing the value and the derivative.  But in every call of ```valueAndDerivative``` a new sample should be chosen. The other behavior can be achieved by passing a sampler to the metric, which always samples the same points.

Issue #175, opened 2016-12-30
GPA procedure changes Reference in DataCollection


marcelluethi:
The reference mesh of a DataCollection is set to the Procrustes mean when performing a GPA. 

Hence the following code returns false
```
val dc :  DataCollection = ???
val dcAfterGPA = DataCollection.gpa(dc)
dc.reference == dcAfterGPA.reference // false
```

I would argue that this is unintuitive and that only the transformations should change.

Andreas-Forster:
I think this is perfectly fine as it is, as for gpa the steps are:

1. choose reference
2. align all shapes to reference
3. compute mean shape
4. if mean and reference are different, set mean as reference and continue with 2.

Hence this algorithm clearly changes the reference. If one does not want to have a new reference one should create a new DataCollection with the aligned data but a different reference.

marcelluethi:
In traditional shape modelling I would agree with you. However, in scalismo (and the Gaussian process framework) we usually regard the reference simply as the domain over which the deformation fields / transformations are defined, and do not assign a semantic meaning to it (such as that the reference is the mean shape). While GPA clearly changes the mean, it does not change the domain over which the deformations are defined.

Andreas-Forster:
I see your point, but then a different name should be introduced for this functionality. I think it is not a good idea to use widely used concepts and then change their behavior slightly without changing the name too. This makes it much more confusing for others to start with.

marcelluethi:
I completely agree. The concept of the DataCollection in general is a bit difficult to grasp in the shape modelling context. Maybe we can simplify it in future versions.

PR #202, opened 2017-04-30
Changed GPA alignement, such that it does not change the reference.

marcelluethi:
As discussed in Issue #175, the GPA alignment should not change the reference surface. 
In this PR we fix this issue and provide an implementation which keeps the reference fixed. 
Furthermore, we introduce two new convenient methods into the ```DataCollection``` class, to obtain the mean transformation and the mean surface represented by the ```DataCollection```.

In case we find a valid use case that requires the change of reference to the mean, we could introduce a method ```DataCollection.changeReference(newReference : TriangleMesh```), which makes this change simple but explicit.

### DIFF
--- a/src/main/scala/scalismo/statisticalmodel/dataset/DataCollection.scala
+++ b/src/main/scala/scalismo/statisticalmodel/dataset/DataCollection.scala
@@ -18,7 +18,7 @@ package scalismo.statisticalmodel.dataset
 import java.io.File
 
 import scalismo.common.{ RealSpace, UnstructuredPointsDomain }
-import scalismo.geometry.{ Point3D, Dim, Point, _3D }
+import scalismo.geometry._
 import scalismo.io.MeshIO
 import scalismo.mesh._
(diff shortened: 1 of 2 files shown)


src/main/scala/scalismo/statisticalmodel/dataset/DataCollection.scala
```diff
@@ -76,6 +75,34 @@ case class DataCollection(reference: TriangleMesh[_3D], dataItems: Seq[DataItem[
...
+  val meanTransformation: Transformation[_3D] = {
+
+    Transformation {
+
+      (pt: Point[_3D]) =>
+        {
+          var meanPoint = Point3D(0, 0, 0)
+          var i = 0
+          while (i < dataItems.size) {
+            meanPoint += dataItems(i).transformation(pt).toVector / dataItems.size
```
Andreas-Forster: Here the division could be pulled out of the while loop.

Ghazi-Bouabene:
Also agree on merging this.

PR #203, opened 2017-05-03
Changed hierarchy of ClosestPoint classes (see Issue #201)

### LINKED ISSUE #201: type hierarchy of ClosestPoint does not represent intent
Currently a result of a closestPoint operation can be either of type ```ClosestPoint```, which indicates that only the closest point and the distance was computed, or one of the subclasses ```ClosestPointIsVertex``` , ```ClosestPointOnLine``` or ```ClosestPointInTriangle```, which also provide access to meta data. 

However, these types should all be distinct and ```ClosestPoint``` should not be the base class, but rather a separate type, in order to make it possible to understand from the type signature, which information is available.

Andreas-Forster:
The hierarchy of `ClosestPoint` and its sub classes is redesigned in this pull request to better reflect the underlying concept. This resolves #201.

New is that the `ClosestPoint` is a class without sub-classes and is used if only the location and or the distance to the closest point on a surface is needed. The `ClosestPoint` on a mesh is returned by the method `closestPoint` from the mesh operations.

The other method in mesh operations `closestPointOnSurface` returns now the abstract class `ClosestPointOnSurface`. The abstract class looks like the `ClosestPoint` but is specialized in three sub-classes: `ClosestPointIsVertex`, `ClosestPointOnLine` and `ClosestPointInTriangle`. These classes contain additional information that might be usefull in some cases. As the super-class `ClosestPointOnSurface` is now a sealed abstract class, pattern matching is simpler and cases of arbitrary classes or the super-class are no longer needed.


marcelluethi:
@Andreas-Forster This looks good to me. Thanks a lot for this improvement.

Issue #204, opened 2017-05-04
Bug in computation of covariance matrix (MultivariateNormalDistribution)


marcelluethi:
One way of creating a MultivariateNormalDistribution is by specifying the mean and three orthonormal directions (with associated variance) (see https://github.com/unibas-gravis/scalismo/blob/master/src/main/scala/scalismo/statisticalmodel/MultivariateNormalDistribution.scala#L159) 

The computation does not compute the correct covariance matrix.

sschoenborn:
Fixed in #205

PR #205, opened 2017-05-04
Corrected computation of covariance matrix in MultivariateNormalDistribution

marcelluethi:
As reported in #204, the covariance matrix was not correctly computed from the principal axis. This PR fixes this bug (and adds a corresponding test)


sschoenborn:
This pull request looks good and prevents nasty bugs which are hard to detect, thanks

PR #207, opened 2017-05-24
added shutdown hook to ensure termination of vtk garbage collector

marcelluethi:
Enabling  the VTK Garbage collector (introduced in version 0.15.1 to avoid memory leaks) sometimes prevents the application from terminating. This is fixed by using a shutdown hook, in which the garbage collector is explicitly stopped.

### DIFF
--- a/src/main/scala/scalismo/package.scala
+++ b/src/main/scala/scalismo/package.scala
@@ -39,26 +39,28 @@ package object scalismo {
       NativeLibraryBundles.initialize(mode)
 
       if (gcInterval > 0) {
-        setupVTKGCThread(gcInterval, TimeUnit.MILLISECONDS)
+        setupVTKGCThread(gcInterval)
       }
       initialized(0) = true
     }
 
   }
 
-  private def setupVTKGCThread(gcInterval: Long, unit: TimeUnit): Unit = {
+  private def setupVTKGCThread(gcInterval: Long): Unit = {
+
+    val runGC = new Runnable() {
+      override def run() {
+        vtkObjectBase.JAVA_OBJECT_MANAGER.gc(false)
+      }
+    }
 
     val gcThread = new Thread {
       override def run() {
         while (true) {
-          unit.sleep(gcInterval)
+          Thread.sleep(gcInterval)
 
           // As vtk is very sensitive to threading issues, we run the gc on the EDT thread.
-          SwingUtilities.invokeLater(new Runnable() {
-            override def run() {
-              vtkObjectBase.JAVA_OBJECT_MANAGER.gc(false)
-            }
-          });
+          SwingUtilities.invokeLater(runGC);
 
         }
       }
--- a/src/main/scala/scalismo/package.scala
+++ b/src/main/scala/scalismo/package.scala
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
-import java.util.concurrent.TimeUnit
+import java.util.concurrent
+import java.util.concurrent.{ Executors, TimeUnit }
+import javax.swing.SwingUtilities
 
 import scalismo.support.nativelibs._
 import vtk.vtkObjectBase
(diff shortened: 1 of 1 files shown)


clangguth:
Told you so :-) https://github.com/unibas-gravis/scalismo/pull/194

That said, I like this approach! :+1: 

It's way better than the approach I took in the UI, so once this is active, the corresponding stuff in the UI can actually be deleted.

marcelluethi:
@clangguth Oops, I missed your comment to my last PR. Otherwise I would have, of course, listened to you :-)
But I am happy to hear that this looks okay.

src/main/scala/scalismo/package.scala
```diff
@@ -38,6 +38,10 @@ package object scalismo {
       gc.SetScheduleTime(60, TimeUnit.SECONDS)
       gc.Start()
 
+      // we need to stop the garbage collection when the JVM shutdowns.
+      // Otherwise it prevents the JVM from shutting down correctly.
+      Runtime.getRuntime().addShutdownHook(new Thread { gc.Stop() })
```
clangguth: I'm not entirely sure if this is the correct way. To my understanding, this calls gc.Stop() in the **initializer** of the Thread, whereas it should actually be called when the Thread is **run**. Shouldn't this rather be

```
    Runtime.getRuntime.addShutdownHook(new Thread(){
      override def run(): Unit = gc.Stop()
    })
```

?

clangguth: PS: From https://docs.oracle.com/javase/7/docs/api/java/lang/Runtime.html#addShutdownHook(java.lang.Thread)

"A shutdown hook is simply an initialized but unstarted thread. When the virtual machine begins its shutdown sequence it will start all registered shutdown hooks in some unspecified order and let them run concurrently"

marcelluethi: I think you are right. But unfortunately once it is done right it does not solve the problem any longer.

clangguth: Huh? That's interesting. That would be an indicator that the shutdown hook doesn't get executed '''because''' the GC is still running? Or something similar...

Either way, I don't really understand how this version could work, then. From my understanding, it would immediately terminate the garbagecollector, and install a no-op shutdown hook.

I think this might require more investigation... I'll take a look at this maybe tomorrow.

marcelluethi: @clangguth The problem seems to be that the garbage collector does not run in a daemon thread and hence is not terminated at the end. Maybe something like this would work? What do you think?
```scala
      val gcThread = new Thread {
        override def run() {
          while (true) {
            Thread.sleep(60 * 1000) // wait for 60 seconds
            val gc = vtkObjectBase.JAVA_OBJECT_MANAGER.gc(false)
          }
        }
      }
      gcThread.setDaemon(true)
      gcThread.start()
```

clangguth: Yeah, I thinks that's probably a sensible and working solution. A few comments though:
- VTK is extremely sensitive to threading issues, so we should mimic the "original" behavior as much as possible. That means running the actual GC in the Swing EDT (runLater(), cf. the source code of the VTK-provided auto-GC implementation).
- should we allow for customizations of the interval and/or debug flag, or is this "YAGNI" territory :-) ?
- what about the (unusual and improbable) case when someone/something else explicitly starts the original VTK auto-GC?

marcelluethi: @clangguth Good points.  We could have an optional argument to initialize, which allows the user to specify the interval. By setting it to 0 we could disable our own garbage collection. In this way, a user who wants to use the original VTK auto-GC could at least disable our GC mechanism. 

I will adapt my code accordingly and update the PR.

clangguth:
Looks good to me, even if I'd personally get rid of the (redundant and distracting) TimeUnit parameter -- it's a private method, so the only value this will ever take is "milliseconds" anyway, but "unit.sleep" is much less understandable  than "Thread.sleep()" (where it's clear that it takes milliseconds). Plus, the latter will be a few nanoseconds more performant...

Speaking of micro-optimizations, you could also define the Runnable once, as a val, instead of re-creating it on every iteration.

Anyhow, once this is merged, remember to throw out the then-redundant stuff from the UI :-)

marcelluethi:
@clangguth micro-optimized according to your suggestions.

Issue #208, opened 2017-05-29
DiscreteDomain: Hash code is broken


sschoenborn:
`DiscreteDomain` calculates its `hashCode` based on the `points` iterator: 
[DiscreteDomain, L73](https://github.com/unibas-gravis/scalismo/blob/develop/src/main/scala/scalismo/common/DiscreteDomain.scala#L73)

```scala
override def hashCode() = points.hashCode()
```

The hash value then depends on the concrete iterator instance rather than the actual point sequence.

Consequence: Meshes (which use `UnstructuredPointsDomain`) do not hash properly

- Equality uses `points.toSeq` to compare - It is thus not affected by this bug

PR #209, opened 2017-05-29
Fixes inconsistent DiscreteDomain.hashCode

sschoenborn:
`DiscereteDomain` calculated `hashCode` of iterator on underlying sequence of points rather than on the sequence itself. This led to random hashCodes even on the same domain instance. See issue #208


marcelluethi:
@sschoenborn nice catch 👍 Thanks a lot for fixing this.

PR #210, opened 2017-05-30
Implicit Random needs now explicit import for the default value

Andreas-Forster:
Having an implicit val in the companion object of `scalismo.utils.Random` prevented the compiler from helping the developers. The compiler did not complained about a missing implicit. This resulted in hardly traceable bugs where seeding the random number generator did not lead to reproducible results.

Existing code can be adapted in three ways to use the new `Random` class:

- You have to do nothing if you defined an implicit val with a seeded `Random` object in your application.
- You have to import `scalismo.utils.Random.implicits._` if you want a random number generator that is seeded with the current time in milliseconds. This corresponds to the behavior of specifying nothing using the old version.
- If your code is a library you have to extend your function with an implicit `Random` parameter. (Or, what we do not recommend, use one of the above two options in order to define an implicit.)

Remark: The additional class `Random` and the renamed class `RandomNumberGenerator` allows us to have at a single location the verbose error message about the missing implicit object.

### DIFF
--- a/src/main/scala/scalismo/utils/Random.scala
+++ b/src/main/scala/scalismo/utils/Random.scala
@@ -28,7 +28,8 @@ import org.apache.commons.math3.random.MersenneTwister
  *
  * @param seed The seed for the Random number generator
  */
-case class Random(seed: Long) {
+
+private[utils] case class RandomNumberGenerator(seed: Long) {
 
   implicit val breezeRandBasis: RandBasis = new RandBasis(new ThreadLocalRandomGenerator(new MersenneTwister(seed)))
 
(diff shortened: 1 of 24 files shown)


src/main/scala/scalismo/utils/Random.scala
```diff
@@ -46,8 +47,27 @@ case class Random(seed: Long) {
 
 }
 
+@scala.annotation.implicitNotFound("missing implicit Random\nTo fix the missing implicit either use:\n\timport scalismo.utils.Random.implicitis._\n...or create a seeded random object:\n\timport scalismo.utils.Random\n\timplicit val rng = Random(1024L)")
```
marcelluethi: there is a typo: should be implicits not implicitis

marcelluethi:
:+1:  I like the proposal a lot, as it definitely helps us to avoid bugs.

src/main/scala/scalismo/utils/Random.scala
```diff
@@ -46,8 +47,27 @@ case class Random(seed: Long) {
...
+  def breezeRandomUnform(a: Double, b: Double) = rng.breezeRandomUniform(a, b)
+}
+
 object Random {
 
-  implicit val randomGenerator = Random(System.nanoTime())
+  def apply(seed: Long) = new Random()(RandomNumberGenerator(seed))
+
+  object implicits {
+    implicit val randomGenerator = new Random()(RandomNumberGenerator(System.currentTimeMillis()))
```
sschoenborn: why change back to `currentTimeMillis`?

src/main/scala/scalismo/utils/Random.scala
```diff
@@ -28,7 +28,8 @@ import org.apache.commons.math3.random.MersenneTwister
  *
  * @param seed The seed for the Random number generator
  */
-case class Random(seed: Long) {
+
+private[utils] case class RandomNumberGenerator(seed: Long) {
 
   implicit val breezeRandBasis: RandBasis = new RandBasis(new ThreadLocalRandomGenerator(new MersenneTwister(seed)))
```
sschoenborn: While you are at it: this should be changed to `RandBasis.withSeed(seed)`, see #212

sschoenborn:
@marcelluethi is this test now failing consistently? should we change the random seed/evaluation criterion?

Andreas-Forster:
Increasing the number of points that are sampled to estimate the new uncertainty fixes the test. But I think the test should be fixed in a more stable way.

marcelluethi:
@Andreas-Forster  I think increasing the number of points to make the estimate more accurate is a good fix to the problem. Performance is not such a big issue here, as the estimation is only done in the rare case where we have a non-rigid transformation that we want to apply to the landmarks.

PR #213, opened 2017-06-22
Improved registration interface.

marcelluethi:
The interface has been changed to be similar to the one used for the Metropolis Hastings algorithm. Further, the metric is now defined for a fixed transformation space, which, due to the stronger assumptions, allows for better performance optimizations. The ```RegistrationConfig``` class is no longer used.

Registration is now performed as follows:
```scala
  val transformationSpace : TransformationSpace[_3D] = ??? 

  // setting up metric
  val fixedImage : ScalarImage[_3D] = ???
  val movingImage : DifferentiableScalarImage[_3D] = ???
  val sampler : Sampler[_3D] = ???
  val metric = MeanSquaresMetric(fixedImage, movingImage, transformationSpace, sampler))

 // setting up the regularizer
  val regularizer = L2Regularizer(transformationSpace)

  // get an iterator to drive the registration
  val initialParameters : DenseVector[Double] = ???

  val registrationIterator = Registration(
        metric,
        regularizer,
        regularizationWeight = 0.0,
        LBFGSOptimizer(maxIter = 300)
      ).iterator(initialParameters)
```


Ghazi-Bouabene:
Making the parameters more prominent in the metric (instead of the transform itself) makes a lot of sense to me, especially given the MCMC usage as you mentioned. I would be ok for merging provided that the maxIter issues are fixed in the tests.

PR #216, opened 2017-08-19
Discrete field with domaintype

marcelluethi:
A DiscreteField is defined using a DiscreteDomain and a set of values, which correspond to the values at the points of the domain. So far the information about the concrete type of the DiscreteDomain was not available
in the DiscreteField. It was only know that it was a subtype of DiscreteField. By introducing the concrete Domain as a type parameter into DiscreteField, i.e. by defining
```
class DiscreteField[D <: Dim, DDomain <: DiscreteDomain[D], A](val domain: DDomain, val data: IndexedSeq[A])
```
we can retain this information. This makes it possible that methods defined on discrete fields can exploit more information about the domain, such as for example that it is defined on a regular grid. A typical example is an interpolation method, which can be much
more efficiently implemented if the underlying domain is an image domain.

### DIFF
--- a/src/main/scala/scalismo/statisticalmodel/DiscreteGaussianProcess.scala
+++ b/src/main/scala/scalismo/statisticalmodel/DiscreteGaussianProcess.scala
@@ -27,7 +27,7 @@ import scalismo.utils.Random
  * While this is technically similar to a MultivariateNormalDistribution, we highlight with this
  * class that we represent (discrete) functions, defined on the given domain.
  */
-class DiscreteGaussianProcess[D <: Dim: NDSpace, Value] private[scalismo] (val mean: DiscreteField[D, Value],
+class DiscreteGaussianProcess[D <: Dim: NDSpace, DDomain <: DiscreteDomain[D], Value] private[scalismo] (val mean: DiscreteField[D, DDomain, Value],
     val cov: DiscreteMatrixValuedPDKernel[D])(implicit val vectorizer: Vectorizer[Value]) {
   self =>
 
@@ -37,7 +37,7 @@ class DiscreteGaussianProcess[D <: Dim: NDSpace, Value] private[scalismo] (val m
 
   val outputDim = vectorizer.dim
 
-  def sample()(implicit rand: Random): DiscreteField[D, Value] = {
+  def sample()(implicit rand: Random): DiscreteField[D, DDomain, Value] = {
     // define the mean and kernel matrix for the given points and construct the
     // corresponding MV Normal distribution, from which we then sample
 
@@ -49,7 +49,7 @@ class DiscreteGaussianProcess[D <: Dim: NDSpace, Value] private[scalismo] (val m
     val sampleVec = mvNormal.sample
 
     // The sample is a vector. We convert it back to a discreteVectorField.
-    DiscreteField.createFromDenseVector[D, Value](domain, sampleVec)
+    DiscreteField.createFromDenseVector[D, DDomain, Value](domain, sampleVec)
   }
 
   /**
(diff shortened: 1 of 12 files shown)


src/main/scala/scalismo/statisticalmodel/DiscreteGaussianProcess.scala
```diff
@@ -63,13 +63,13 @@ class DiscreteGaussianProcess[D <: Dim: NDSpace, Value] private[scalismo] (val m
    * The marginal distribution for the points specified by the given point ids.
    * Note that this is again a DiscreteGaussianProcess.
    */
-  def marginal(pointIds: Seq[PointId])(implicit domainCreator: UnstructuredPointsDomain.Create[D]): DiscreteGaussianProcess[D, Value] = {
+  def marginal(pointIds: Seq[PointId])(implicit domainCreator: UnstructuredPointsDomain.Create[D]): DiscreteGaussianProcess[D, UnstructuredPointsDomain[D], Value] = {
```
marcelluethi: Here we should use DDomain instead.

src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
```diff
@@ -164,25 +164,25 @@ case class DiscreteLowRankGaussianProcess[D <: Dim: NDSpace, Value] private[scal
    * data are defined by the pointId. The returned posterior process is defined at the same points.
    *
    */
-  def posterior(trainingData: IndexedSeq[(PointId, Value, MultivariateNormalDistribution)]): DiscreteLowRankGaussianProcess[D, Value] = {
+  def posterior(trainingData: IndexedSeq[(PointId, Value, MultivariateNormalDistribution)]): DiscreteLowRankGaussianProcess[D, DDomain, Value] = {
     DiscreteLowRankGaussianProcess.regression(this, trainingData)
   }
 
-  override def marginal(pointIds: Seq[PointId])(implicit domainCreator: UnstructuredPointsDomain.Create[D]): DiscreteLowRankGaussianProcess[D, Value] = {
+  override def marginal(pointIds: Seq[PointId])(implicit domainCreator: UnstructuredPointsDomain.Create[D]): DiscreteLowRankGaussianProcess[D, UnstructuredPointsDomain[D], Value] = {
```
marcelluethi: here we should use a more general implementation of marginal, which takes a generic DDomain

src/main/scala/scalismo/statisticalmodel/DiscreteGaussianProcess.scala
Ghazi-Bouabene: Not so sure, even if the initial DDomain might be a grid, the user might choose to marginalize over a set of points that are not necessarily a grid.

Ghazi-Bouabene:
I don't quite get the need for this change yet.
Currently, to interpolate and image, we use image.interpolate .. meaning that we already know that the domain is a DiscreteImageDomain and hence can write optimized code for this. To me, this change suggest that you want to be able to write optimized code in generic classes. Or?

src/main/scala/scalismo/statisticalmodel/DiscreteGaussianProcess.scala
marcelluethi: Good point. The method should actually have its own domain type.

marcelluethi:
@Ghazi-Bouabene The main motivation for the change is that I would like to be able to interpolate a DiscreteGaussianProcess. As in the current implementation of scalismo, the domain type is not part of the definition of a DiscreteField (it is only known that the domain is a subclass of DiscreteField) we cannot use any other interpolator than the NearestNeighborInterpolator to interpolate GPs. With this change, we can use any user defined interpolation strategy that works on the right domain type.

Ghazi-Bouabene:
@marcelluethi Yes that's what I imagined from PR #217, where you can create different types of interpolators, some more optimized for certain domains than others and pass them in to the interpolate method ..  But then again, the type of the domain should already be known in the interpolator (that might be a dummy nearest neighbor one, or an image domain interpolator for example). For example, if the interpolate method of a DiscreteLowRankGaussianProcess[D <: Dim: NDSpace, DDomain <: DiscreteDomain[D], Value] expects an Interpolator[D, DDomain, Value], why should the DiscreteFields as well be parametrized on the type. We can safely assume that interpolator.interpolate(..) will receive a DiscreteField that is defined on a discerete domain of the correct type, or?

Ghazi-Bouabene:
:+1:

src/main/scala/scalismo/statisticalmodel/DiscreteGaussianProcess.scala
marcelluethi: @Ghazi-Bouabene In contrast to what I previously said, It seems to be correct to use only a seq of pointIds here and not a domain. We are already in a discrete setting here.

Ghazi-Bouabene: @marcelluethi Does this mean that when marginalizing a DiscreteLowRankGaussianProcess[_3D, DiscreteImageDomain[_3D], Double] we would always end up with DiscreteLowRankGaussianProcess[_3D, UnustructuredPointsDomain[_3D], Double] ? An efficient interpolation of the marginal later on would not be possible.

marcelluethi: It is at least not straight-forward to implement. As the GP is already discrete we cannot sample it at arbitrary domain points. So if the GP represents for example a ```UnstructuredPointsDomain```, it is not possible to resample it with a ```DiscreteImageDomain```. Encoding this in types seems not worth the trouble. But we can achieve the same functionality by first interpolating and then discretizing again.

Ghazi-Bouabene: @marcelluethi Good point.

PR #217, opened 2017-08-19
Interpolation mechanism for discrete fields

marcelluethi:
*This PR depends on #216, which should be merged before this one is considered.*

A new interpolation mechanism for discrete fields is introduced. Compared to the original implementation, the new method takes an interpolator object with the interpolation algorithm. We can now write:
```
val df : DiscreteField[D, Domain, A] = ???
val interpolator = NearestNeighborInterpolator()
val field : Field[D, A] = df.interpolate(interpolator)
```

This makes it possible to use different interoplation algorithms using the same interface and separates the interpolation algorithm from the field implementation. The same mechanism is also used for the interpolation of a DiscreteLowRankGaussianProcess, i.e. we can write
```
val discreteGp : DiscreteLowRankGaussianProcess[D, Domain, A]
val interpolator = NearestNeighborInterpolator()
val gp : LowRankGaussianProcess[D, A] = discreteGp.interpolate(interpolator)
```

### DIFF
--- a/src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
+++ b/src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
@@ -43,8 +43,8 @@ import scalismo.utils.{ Memoize, Random }
  * @see [[DiscreteLowRankGaussianProcess]]
  */
 
-case class DiscreteLowRankGaussianProcess[D <: Dim: NDSpace, Value] private[scalismo] (_domain: DiscreteDomain[D], meanVector: DenseVector[Double], variance: DenseVector[Double], basisMatrix: DenseMatrix[Double])(override implicit val vectorizer: Vectorizer[Value])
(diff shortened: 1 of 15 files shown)


src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
```diff
@@ -186,14 +187,31 @@ case class DiscreteLowRankGaussianProcess[D <: Dim: NDSpace, DDomain <: Discrete
...
+   */
+  def interpolate(interpolator: FieldInterpolator[D, DDomain, Value]): LowRankGaussianProcess[D, Value] = {
+    val newKLBasis = for (DiscreteEigenpair(eigenVal, eigenFun) <- klBasis) yield {
+      Eigenpair(eigenVal, eigenFun.interpolate(interpolator))
+    }
+    val newMean = this.mean.interpolate(interpolator)
+
+    new InterpolatedLowRankGaussianProcess(Field(RealSpace[D], newMean), newKLBasis, this, interpolator)
+
+    new LowRankGaussianProcess[D, Value](newMean, newKLBasis)
```
marcelluethi: This line should be removed

Ghazi-Bouabene:
:+1:

PR #218, opened 2017-08-20
Moved mesh.interpolator to common.interpolation

marcelluethi:
As ```scalismo.mesh.interpolator``` is of more common interest than only for interpolating values defined on meshes, it was moved to common.interpolation and its name was changed to ```ValueInterpolator``` such that it is not confused with more complex interpolators such as image interpolators..

Other minor changes:
* The coefficients for defining the convex combination were changed
  from Float to Double
* Introduced convenience method for obtaining the implict instance (via apply)

### DIFF
--- a/src/main/scala/scalismo/common/interpolation/ValueInterpolator.scala
+++ b/src/main/scala/scalismo/common/interpolation/ValueInterpolator.scala
@@ -13,20 +13,20 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package scalismo.mesh
+package scalismo.common.interpolation
 
-import scalismo.geometry._
 import scalismo.geometry.Vector._
+import scalismo.geometry._
 
 /** defines (convex) interpolation between two values */
-trait Interpolator[@specialized(Double, Float) A] {
+trait ValueInterpolator[@specialized(Double, Float) A] {
   def blend(obj1: A, obj2: A, l: Double): A
 
   /** mix multiple values, forms convex combination (with normalized weights) */
   // loop: better performance than list eating tailrec
-  def convexCombination(first: (A, Float), rest: (A, Float)*): A = {
+  def convexCombination(first: (A, Double), rest: (A, Double)*): A = {
     var mix: A = first._1
-    var f: Float = first._2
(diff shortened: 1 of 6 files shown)


src/main/scala/scalismo/common/interpolation/ValueInterpolator.scala
```diff
@@ -51,41 +51,44 @@ trait Interpolator[@specialized(Double, Float) A] {
     var mix: A = first
     var n = 1
     for (next <- rest) {
-      mix = blend(mix, next, n.toFloat / (n + 1))
+      mix = blend(mix, next, n / (n + 1))
```
Andreas-Forster: The conversion can not be ommited here. I'm quite sure it will end up as hinterher division, which is not what we want here.

Andreas-Forster: With passing Doubles to the convex combination method i agree.

Why should we not stay with the simpe name for the simple interpolator? Calling the image interpolator ImageInterpolator does not lead to name clashes and seems to me to be the more natural solution.

Further would the numerics not be a more obvious package for this class than common?

marcelluethi: @Andreas-Forster Nice catch. I mistakenly thought the toFloat was because of a Double to Float conversion.
I also completely agree that the ValueInterpolator should go into the numerics package. 

Regarding the naming: The other interpolators do already have more specific names (e.g. FieldInterpolator, LinearImageInterpolator, etc.) to make it more clear what their purpose is. Leaving the name of this class as Interpolator would suggest that it is somehow more general than all the others. This is clearly not true, as this interpolator has  the very narrow purpose of interpolating two values. Furthermore it will never be called directly by a user. Therefore I think keeping the current name would be confusing .

Renaming the Interpolator trait should not break any code, as this class is not used directly from user code. Even if it was, it is a good moment to rename it, as moving it to another package would anyway break the code.

Ghazi-Bouabene:
:+1: ..  also agree on the move to numerics package (together with the Integrator)

PR #220, opened 2017-08-27
Made DiscreteGp covariant on DiscreteDomain

marcelluethi:
The DiscreteGaussianProcess and DiscreteLowRankGaussianProcess were so far invariant in the domain argument. In line with the DiscreteField, it should, however be covariant, such that we can pass a ```DiscreteLowRankGP[D, DiscreteImageDomain[D], V]``` where a ```DiscreteLowRankGP[D, DiscreteDomain[D], V]``` is expected.


Andreas-Forster:
While reflecting only a limited amount of time about it, it makes totally sense that we can pass a DLRGP with stricter domain to a method expecting only a DLRGP with a parent type of the stricter domain. Maybe if the inheritance of the domain is abused this might lead to strange result. But I think we should not argue with this kind of problems.

So i agree to merge this PR.

PR #221, opened 2017-08-31
Linear Interpolator

thogerig:
Linear Interpolator for 1D, 2D and 3D images including tests.

### DIFF
--- a/src/main/scala/scalismo/common/interpolation/LinearInterpolation.scala
+++ b/src/main/scala/scalismo/common/interpolation/LinearInterpolation.scala
@@ -23,9 +23,9 @@ import scalismo.numerics.ValueInterpolator
 
 trait LinearImageInterpolator[D <: Dim, A] extends FieldInterpolator[D, DiscreteImageDomain[D], A] {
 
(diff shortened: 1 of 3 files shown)


src/main/scala/scalismo/common/interpolation/LinearInterpolation.scala
```diff
@@ -0,0 +1,176 @@
...
+package scalismo.common.interpolation
+
+import scalismo.common.{ DiscreteField, Field }
+import scalismo.geometry.{ Dim, IntVector, IntVector2D, IntVector3D, NDSpace, Point, _1D, _2D, _3D }
+import scalismo.image.DiscreteImageDomain
+import scalismo.numerics.ValueInterpolator
+
+trait LinearImageInterpolator[D <: Dim, A] extends FieldInterpolator[D, DiscreteImageDomain[D], A] {
+
+  implicit def ndSpace: NDSpace[D]
```
marcelluethi: the two implicits should be made protected and in the concrete subclasses declared as private.

thogerig:
Marcel, thanks for the comment. The two defs are now protected.

PR #225, opened 2017-12-02
Added new registration metric SumOfPointwiseLossMetric.

marcelluethi:
The new metric makes it easy to implement different metrics that use robust loss functions. 
As a concrete (and very useful) implementation, the Huber loss function is provided with this PR. 

Since the need for a robust loss function is arising so often in practice, I would argue that this feature should be part of the 0.16 release of Scalismo (in spite of that the first release candidate has already been released). Inclusion does not pose a risk, as it does not touch existing functionality. Furthermore, the metric has already been tested outside of scalismo.

### DIFF
--- a/src/main/scala/scalismo/registration/PointwiseHuberLossMetric.scala
+++ b/src/main/scala/scalismo/registration/PointwiseHuberLossMetric.scala
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2015 University of Basel, Graphics and Vision Research Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package scalismo.registration
+
+import scalismo.geometry.{ Dim, NDSpace }
+import scalismo.image.{ DifferentiableScalarImage, ScalarImage }
+import scalismo.numerics._
+import scalismo.utils.Random
+
+/**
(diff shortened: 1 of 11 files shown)


src/main/scala/scalismo/registration/PointwiseHuberLossMetric.scala
```diff
@@ -0,0 +1,49 @@
...
+ * values below this threshold and linearly for values above this threshold.
+ * @see SumOfPointwiseLossMetric.
+ *
+ */
+case class PointwiseHuberLossMetric[D <: Dim: NDSpace](fixedImage: ScalarImage[D],
+  movingImage: DifferentiableScalarImage[D],
+  transformationSpace: TransformationSpace[D],
+  sampler: Sampler[D],
+  delta: Double = 1.345)(implicit rng: Random)
+    extends SumOfPointwiseLossMetric(fixedImage, movingImage, transformationSpace, sampler) {
```
Ghazi-Bouabene: Maybe the MeansquaresMetric should also be brought under this hierarchy then?

marcelluethi: If we would define as the loss function the squared function, we would get the sum of squared differences metric. This is slightly different from the mean squares metric, which takes the average over all the points.

But you are raising a very good point. Maybe we should also average the loss in the new metric, as otherwise the value of the loss function depends on the number of points that are sampled? This would mean that the more points we sample, the larger the weight of the metric compared to the regularizer.

Ghazi-Bouabene: That would make a lot of sense, yes. In which case, the Meansquares would also be under that hierarchy.

marcelluethi: I just realized that I was wrong. The averaging is already done by the integrator. Strangely enough, the integrator takes the number of points, but not the volume into account. This might actually be a bug in the intergrator ...

I will put the mean squares metric under this scheme here and also check what should be the correct behavior of the integrator.

marcelluethi:
@Ghazi-Bouabene The commends should now all be addressed. The mean squares metric is now a special case of the MeanPointwiseLossMetric. Offline tests on real data showed that the registration still works and gives the same results as the old metric.

Issue #238, opened 2018-04-29
Similarity transform for statistical models


madsendennis:
As of now SSM only allows for rigid transformations. The suggestion is to extend to also allow for similarity transform. At first, perform a rigid transform and then a scaling transform for the basis vectors according to the computed scaling factor between original and target landmarks (as done in similarity3DLandmarkRegistration).
Reference mesh and mean deformation field should also be updated.

PR #239, opened 2018-05-01
Added similarity tranform to SSM

madsendennis:
Fix for issue #238. 
Added SMM similarity transform and merged the rigid transform to avoid duplicate.
Simple test added to check back and forth similarity transform.


Ghazi-Bouabene:
In most of our applications, we use a definition of shape that includes scale since we are interested in modelling it as well (e.g long vs short femurs).

By allowing to transform a model while changing its scale, one might break the assumptions that a model does on the scale: one could end up having an extremely long femur (instance of the model) being considered as highly likely acording to the transformed model.

Since this method is a convenience one and, given that it could lead other users into faulty usage, I'd rather reject this PR.

Issue #235, opened 2018-04-20
Move VertexColorMesh3D from scalismo-faces to scalismo


Ghazi-Bouabene:
This class is defined here: 
https://github.com/unibas-gravis/scalismo-faces/blob/master/src/main/scala/scalismo/faces/mesh/Mesh.scala#L43

The move would require to also move the RGBA class. The advantage would be that we could release [this change](https://github.com/marcelluethi/scalismo-ui-1/commit/22d15b7b13ad542188644ad63385e412c0e474dc) to the official scalismo-ui to visualize vertex colored meshes, without the dependency on scalismo-faces.

BernhardEgger:
I'm not sure if this solves the problem in a sustainable way. Later we would perhaps like to also visualize a Textured Mesh, a Morphable Model (with color model) or a 2D PixelImage. Consequently those would then also go to scalismo? All of this would then drag along more and more stuff from faces needed to visualize this.

Andreas-Forster:
I get your point, but the other options are to make scalismo-ui dependent on scalismo-faces or implement the same thing a second time. I am not sure if one of these is a better option? I also do not see that it is a real consequence that textured meshes and the morphable model with PixelImages will directly follow the vertex color mesh.

I see clearly the need for a vertex color mesh in scalismo-ui and see also the "fear" that moving it could mean scalismo soaks up scalismo-faces nearly as a whole.

BernhardEgger:
It is not really a fear - I anyway almost always use both at the same time :)

What is exactly the problem of scalismo-ui beeing dependent on scalismo-faces?

madsendennis:
Couldn't it quickly become a bit messy as Scalismo-UI and Scalismo-faces both depends on Scalismo. If Scalismo-UI also depends on Scalismo-faces it could end up with having to make a combined release of all 3? Or is this already done today?

Ghazi-Bouabene:
The releases would be too "tied". Say we have a new concept in Scalismo (affecting only the medical area) and we would like to visualize it in the UI .. we would need to first roll out a new scalismo-faces version (even though there are no changes) in order to be able to release a new scalismo-ui version.

Ghazi-Bouabene:
I agree with your argument Bernhard that this could gradually lead to wanting all of Scalismo faces into Scalismo in order to visualize more and more face-related things. 

[...]

PR #244, opened 2018-10-25
moved VertexColorMesh3D from scalismo-faces to scalismo

Ghazi-Bouabene:
(and needed classes along)

This is a followup on PR #243 and issue #235, only intended for merge this time.

### DIFF
--- a/src/main/scala/scalismo/color/ColorSpaceOperations.scala
+++ b/src/main/scala/scalismo/color/ColorSpaceOperations.scala
@@ -1,0 +1,142 @@
+/*
+ * Copyright 2016 University of Basel, Graphics and Vision Research Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


marcelluethi:
@Ghazi-Bouabene In principle I have no objections. But it does not seem super useful without the readers. Do you have any thoughts on how we handle the IO problem?

Ghazi-Bouabene:
I just committed a proposal for the ply io.

marcelluethi:
@Ghazi-Bouabene looks good to me. 
@Andreas-Forster This method will lead to the situation, that the user of scalismo-faces has to choose between two different ply readers. What is your opinion on that? From a Scalismo-users point of view, I would find this solution a pragmatic first step, before we attempt a more complete unification.

src/main/scala/scalismo/io/MeshIO.scala
```diff
@@ -275,6 +356,59 @@ object MeshIO {
...
+    val vtkPd = plyReader.GetOutput()
+
+    val mesh = for {
+      meshGeometry <- MeshConversion.vtkPolyDataToTriangleMesh(vtkPd)
+    } yield {
+      getColorArray(vtkPd) match {
+        case Some(("RGBA", colorArray)) => {
+          val colors = for (i <- 0 until colorArray.GetNumberOfTuples()) yield {
+            val rgba = colorArray.GetTuple4(i)
+            RGBA(rgba(0) / 255.0, rgba(1) / 255.0, rgba(2) / 255.0, rgba(3))
```
Andreas-Forster: is here alpha not from 0..255?

Ghazi-Bouabene:
Should be fixed. Thanks.  Regarding whether or not this should go to a release and testing:  should we do a release candidate? This would make it easier to integrate it and test it in scalismo-ui and scalismo-faces.

PR #247, opened 2018-12-23
fixed most compiler warnings

marcelluethi:
Fixed many small problems concerning unused variables and imports.
A few deprecation warnings still exists, which will need to be fixed in a second pass. 
@Andreas-Forster there are a few unused test functions in MeshBoundaryTests. If we use them as tests, the tests fail. Maybe you can decide what to do with them.

### DIFF
--- a/src/main/scala/scalismo/color/RGB.scala
+++ b/src/main/scala/scalismo/color/RGB.scala
@@ -194,7 +194,7 @@ object RGB {
     }
   }
 
-  private def toInt8(value: Double): Int = (value * 255.0).toInt
+  //private def toInt8(value: Double): Int = (value * 255.0).toInt
 
   private def fromInt8(intValue: Int): Double = intValue / 255.0
 }
--- a/src/main/scala/scalismo/color/RGBA.scala
+++ b/src/main/scala/scalismo/color/RGBA.scala
@@ -205,7 +205,7 @@ object RGBA {
     }
   }
 
-  private def toInt8(value: Double): Int = (value * 255.0).toInt
+  //private def toInt8(value: Double): Int = (value * 255.0).toInt
 
   private def fromInt8(intValue: Int): Double = intValue / 255.0
 }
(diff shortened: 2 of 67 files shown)


src/main/scala/scalismo/color/RGB.scala
```diff
@@ -194,7 +194,7 @@ object RGB {
     }
   }
 
-  private def toInt8(value: Double): Int = (value * 255.0).toInt
+  //private def toInt8(value: Double): Int = (value * 255.0).toInt
```
Andreas-Forster: I think we should not comment out functions to get rid of a compiler warning. Either we want to have the function because we develop a library and not an application so we might provide more functions as we are currently using in the project. We argue that the library can change, or the user might want to code package injections and we have forward-/backward-mappings in. Or we say we want only things we currently need.

src/main/scala/scalismo/color/RGBA.scala
```diff
@@ -205,7 +205,7 @@ object RGBA {
     }
   }
 
-  private def toInt8(value: Double): Int = (value * 255.0).toInt
+  //private def toInt8(value: Double): Int = (value * 255.0).toInt
```
Andreas-Forster: Same as above: "I think we should not comment out functions to get rid of a compiler warning. Either we want to have the function because we develop a library and not an application so we might provide more functions as we are currently using in the project. We argue that the library can change, or the user might want to code package injections and we have forward-/backward-mappings in. Or we say we want only things we currently need."

src/test/scala/scalismo/utils/MemoizationTests.scala
```diff
@@ -28,14 +26,14 @@ class MemoizationTests extends ScalismoTestSuite {
 
     it("yields the same results after memoizing") {
       val testFunMemoized = Memoize(testFun, 100000)
-      for (x <- -2 * Math.PI until 2 * Math.PI by 0.1)
-        testFun(x) should be(testFunMemoized(x))
+      for (x <- BigDecimal(-2.0 * Math.PI) until (2.0 * Math.PI) by 0.1)
```
Andreas-Forster: Do we really need BigDecimal or could we cast the constant Math.PI to double instead? For me, this would be the more natural way.

marcelluethi: This is a deprecation from the scala side. If we want a step-size, which is fractional, we need a BigDecimal (at least this is what the compiler tells you)

Issue #249, opened 2019-01-03
StatismoIO is an unintuitive name for people not familiar with Statismo


marcelluethi:
The objects and methods to read/write statistical mesh models in Scalismo are named after Statismo. This can be confusing for the following reasons:
1. For people unfamiliar with Statismo will at best be seen as an weird abbreviation of statistical model
2. Once we add io methods for other types of statistical models apart from mesh models (such as e.g. deformation models), the name statismo does not distinguish these conepts. 

I propose instead to name the IO object ```StatisticalModelIO``` and name the methods 
```readStatisticalMeshModel```, ```readStatisticalDeformationModel```, etc.

For the transition period, we can keep the existing ```IO``` methods and mark them as deprecated.

PR #250, opened 2019-01-05
Introduced StatisticalModelIO object

marcelluethi:
This PR introduced a new object ```StatisticalModelIO```. This object complements the ```StatismoIO``` class, which provides the concrete implementation of the Statismo file format. The problem with using `StatismoIO``` directly is, that it refers to a concrete implementation of a file format (the statismo file format) rather than the more abstract concept of reading/writing a StatisticalModel. Furthermore, for users who do not know that statismo is a software and fileformat for statistical modelling, the name is unintuitive. 

The ```StatismoIO``` object remains public and is not deprecated, as sometimes we would like to be able to use specific features of the ```StatismoIO``` format, such as the ability to store several models into the same file or read a model catalog. 

This PR addresses issue #249

### DIFF
--- a/src/main/scala/scalismo/io/StatisticalModelIO.scala
+++ b/src/main/scala/scalismo/io/StatisticalModelIO.scala
@@ -19,7 +19,6 @@ import java.io.File
 
 import scalismo.common.{ PointId, UnstructuredPointsDomain }
 import scalismo.geometry.{ Point, _3D }
-import scalismo.io.StatismoIO.StatismoModelType.StatismoModelType
 import scalismo.mesh.{ TriangleCell, TriangleList, TriangleMesh, TriangleMesh3D }
 import scalismo.mesh.TriangleMesh._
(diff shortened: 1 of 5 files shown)


src/main/scala/scalismo/io/StatisticalModelIO.scala
```diff
@@ -19,24 +19,58 @@ import java.io.File
...
+
 import ncsa.hdf.`object`._
 import java.io.DataOutputStream
 import java.io.FileOutputStream
 import java.io.DataInputStream
 import java.io.FileInputStream
 
+import scalismo.io.StatismoIO.StatismoModelType.StatismoModelType
+
+object StatisticalModelIO {
```
Andreas-Forster: I agree that `StatismoIO` is not the most straight forward name to use when regarding the rest of the naming of the library. Is there a reason to not call this `StatisticalMeshModelIO` with the methods `read` and `write`? For me, this would a more straight forward naming scheme.

PR #251, opened 2019-01-06
Abolish dim bound

marcelluethi:
*Attention: Merge only after PR #247 is merged*

Most classes in Scalismo have a type parameter D, which indicates the dimensionality of the class (i.e. _1D, _2D, _3D). This parameter is always explicitly bounded by Dim (i.e we write ```class A[D <: Dim]```). The type bound here is not necessary. While it might make the intention slightly more clear, it also leads to long type signature, which are hard to read. 

This PR proposes to remove this type bound.


marcelluethi:
@Ghazi-Bouabene @Andreas-Forster  Note that this PR is only for discussion. In the meantime it generates many merge conflicts. If agreed upon, replacing of the Dim bound is trivial and will be done in a new PR.

Issue #252, opened 2019-01-10
Mesh metric distance function update from closest point to closestPointOnSurface


madsendennis:
The Average distance and Hausdorff distance functions to measure the distance between two meshes, both use the closest point in pointset function. The distance should instead be taken to the closest point on the surface. 
It could happen that there is 0 distance to the surface on the other mesh, whereas the vertices are not perfectly aligned, which would show as a distance > 0.

Andreas-Forster:
Maybe we should discuss the issue in more detail. I think t#253 his is just one (minor) step toward a better metric. Of course, it is maybe a necessary step towards a better metric.

For me, the metric still assumes two properties that the user might not be aware of. The idea that averaging the metric over the vertices of the mesh [m1](https://github.com/unibas-gravis/scalismo/blob/b8bdde5ee68ff931212f609665d7e36f213dba53/src/main/scala/scalismo/mesh/MeshMetrics.scala#L35) gives you a good approximation of the surface distance has at least two assumptions that are still not fixed after the PR #253 :

- It could be that the m1 mesh vertices are on your mesh surface m2 but the vertices of the m2 mesh are far off.

- The vertices do not necessarily need to have to be equally distributed over the surface.

Both assumptions are crucial that you get a good estimate. The first one can be circumvented by evaluating the measure twice with flipped arguments. The second would need to sample the points on the surface over which the metric is evaluated. So we should agree if we want to aim for a perfect metric or a fast one or even offer different metrics with different tradeoffs between the two extremes.

marcelluethi:
@madsendennis What is the state of this issue? Can we close it or is there a specific reason this is still open?

madsendennis:
The initial mentioned problem with the avgDistance and hausdorffDistance functions have been merged. I'll therefore close this PR. @Andreas-Forster comments regarding a better metric have however not been solved.

PR #253, opened 2019-01-10
Mesh metric distance measure update for avg and hausdorff distance

madsendennis:
This pull request fixes issue: #252


src/test/scala/scalismo/mesh/MeshMetricsTests.scala
```diff
@@ -51,7 +51,7 @@ class MeshMetricsTests extends ScalismoTestSuite {
 
     it("should be (slightly) lower than the average translation applied to each vertex") {
       MeshMetrics.avgDistance(mesh, translatedMesh) should be < translationLength.toDouble
-      MeshMetrics.avgDistance(mesh, translatedMesh) should be(translationLength.toDouble +- (translationLength * 0.2))
+      MeshMetrics.avgDistance(mesh, translatedMesh) should be(translationLength.toDouble +- (translationLength * 0.3))
```
marcelluethi: I don't understand why the margin needs to be increased here. Shouldn't the distance between the meshes now become smaller and not larger, because the distance to the closest point on the vertex is always at least as long as the distance to the closest point on the surface?

marcelluethi: @Andreas-Forster @madsendennis What is the status of this PR? Can it be merged and used in release-0.17 or shall we wait until the next release for a more complete solution?

Andreas-Forster: I think this PR makes sense and improves the current state. So yes I would argue for merging this PR. The "abuse" of the discussion on the issue #252 is more meant as an effort to continue to work towards a metric for users that are not aware of the influence of the discretization of the meshes.

PR #263, opened 2019-02-18
Fixes #262 closestPointOnDistance

### LINKED ISSUE #262: ClosestPointOnSurface does not handle degenerated triangles correctly
When using the function `TriangleMesh3D.operations.closestPointOnSurface` on a triangle mesh with a degenerated triangle, i.e. a triangle with three colinear points, the returned closest point is not always correct. This comes from the fact that calculations used to construct the bounding sphere based index structure and others during a query result in NaN values.

The current version of the code assumes that no degenerated triangles are present in a mesh. Especially the case of a triangle with only colinear points leads to the above-described behaviour. Such triangles are easily produced when using VTK on a volume image with the steps thresholding, contour-filtering, smoothing, decimating.

Here we should introduce additional safety even that we will trade off a bit of performance.

Andreas-Forster:
This PR fixes #262 by testing for the different cases a triangle can be degenerated (co-linear points or single point only). The cases are then handled in the methods involved in the closest point on surface calculations.

The three main changes are:

- The `scalismo.mesh.boundingSpheres.Triangle` checks for the two degeneration cases of co-linear points, and one single location for all three corners (see [here](https://github.com/unibas-gravis/scalismo/compare/develop...Andreas-Forster:fixClosestPointOnDistance?expand=1#diff-ddf9b36802a00ec944b5eb97399f0fbdR31)).

- The calculation of barycentric coordinates handles the different degenration cases (see [here](https://github.com/unibas-gravis/scalismo/compare/develop...Andreas-Forster:fixClosestPointOnDistance?expand=1#diff-ddf9b36802a00ec944b5eb97399f0fbdR58)).

- When generating the bounding sphere `Sphere` the different degenerated cases do not lead to `NaN`'s any more (see [here](https://github.com/unibas-gravis/scalismo/compare/develop...Andreas-Forster:fixClosestPointOnDistance?expand=1#diff-0866c75747cdecadefbdca9ad6ebb86eR369)).

Furthet changes are:

- The `Triangle` class now calculates internally the values based on the corners, so the constructor takes the minimal set of values needed.

- Tests were added (starting from [here](https://github.com/unibas-gravis/scalismo/compare/develop...Andreas-Forster:fixClosestPointOnDistance?expand=1#diff-ff3c5f5658f2e7d661ffd9bce4e1d414R218)) to check that barycnetric coordinates and the bounding sphere are calculated correctly from also degenerated triangles. In addition the test class is moved to the appropriate folder according to the specified package.

### DIFF
--- a/src/main/scala/scalismo/mesh/boundingSpheres/BSDistance.scala
+++ b/src/main/scala/scalismo/mesh/boundingSpheres/BSDistance.scala
@@ -22,7 +22,23 @@ import scalismo.mesh.boundingSpheres.ClosestPointType._
 /**
  * Holds triangles and precalculated vectors.
  */
-private[mesh] case class Triangle(a: EuclideanVector[_3D], b: EuclideanVector[_3D], c: EuclideanVector[_3D], ab: EuclideanVector[_3D], ac: EuclideanVector[_3D], n: EuclideanVector[_3D])
+private[mesh] case class Triangle(a: EuclideanVector[_3D], b: EuclideanVector[_3D], c: EuclideanVector[_3D]) {
+  val ab = b - a
+  val ac = c - a
+  val bc = c - b
+  val n = ab.crossproduct(ac)
+
+  val degenerated = if ( n.norm == 0.0 ) { if (a==b&&b==c) 2 else 1 } else 0
+  // 0: ab, 1: ac, 2: bc
+  val longestSide = {
+    val bc = c - b
+    if ( ab.norm2 > ac.norm2 ) {
+      if ( ab.norm2 > bc.norm2 ) 0 else 2
+    } else {
+      if ( ac.norm2 > bc.norm2 ) 1 else 2
+    }
+  }
+}
 
 /**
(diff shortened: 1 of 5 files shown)


src/main/scala/scalismo/mesh/boundingSpheres/BSDistance.scala
```diff
@@ -22,7 +22,23 @@ import scalismo.mesh.boundingSpheres.ClosestPointType._
 /**
  * Holds triangles and precalculated vectors.
  */
-private[mesh] case class Triangle(a: EuclideanVector[_3D], b: EuclideanVector[_3D], c: EuclideanVector[_3D], ab: EuclideanVector[_3D], ac: EuclideanVector[_3D], n: EuclideanVector[_3D])
+private[mesh] case class Triangle(a: EuclideanVector[_3D], b: EuclideanVector[_3D], c: EuclideanVector[_3D]) {
+  val ab = b - a
```
marcelluethi: These could be made private (though it does not matter much as the class is private and used only locally)

PR #266, opened 2019-03-18
Improved efficiency of SampleCovarianceKernel

marcelluethi:
This PR improves the efficiency of the sample covariance kernel by approximately a factor of 3, by avoiding many unnessecary conversions to/from breeze matrices. 

More importantly, it also greatly reduces the memory overhead.

I propose to treat this as a hotfix and to have it part of the 0.17.1 release.

### DIFF
--- a/src/main/scala/scalismo/kernels/StandardKernels.scala
+++ b/src/main/scala/scalismo/kernels/StandardKernels.scala
@@ -35,36 +35,54 @@ case class GaussianKernel[D](sigma: Double) extends PDKernel[D] {
 
 case class SampleCovarianceKernel[D: NDSpace](ts: IndexedSeq[Transformation[D]], cacheSizeHint: Int = 100000) extends MatrixValuedPDKernel[D] {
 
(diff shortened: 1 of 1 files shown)


src/main/scala/scalismo/kernels/StandardKernels.scala
```diff
@@ -35,36 +35,54 @@ case class GaussianKernel[D](sigma: Double) extends PDKernel[D] {
...
+    for (t <- ts_memoized) {
+      val tx = t(x)
+      val ty = t(y)
+
+      var i = 0
+      while (i < outputDim) {
+        var j = 0
+
+        while (j < outputDim) {
+          ms(i, j) = ms(i, j) + (tx(i) - x(i) - mux(i)) * (ty(j) - y(j) - muy(j))
```
Andreas-Forster: isn't there a transpose missing for (ty...muy(j))**.t** ?

marcelluethi: I think it is correct. Here we just compute the outer product element-wise. I added a commit to make it more clear what is happening. 

There is also a test, which suggests that it is correct.

PR #267, opened 2019-04-04
DataCollection Mesh Loading Memory Fix

thogerig:
This pull request reduces the memory footprint of the function that loads meshes from directory for the data collection. 

This fix relates to the discussion in the Scalismo google group: https://groups.google.com/forum/#!topic/scalismo/MddIEgAnWBg

This should be included as a hotfix.

### DIFF
--- a/src/main/scala/scalismo/statisticalmodel/dataset/DataCollection.scala
+++ b/src/main/scala/scalismo/statisticalmodel/dataset/DataCollection.scala
@@ -128,7 +128,10 @@ object DataCollection {
    */
   def fromMeshDirectory(referenceMesh: TriangleMesh[_3D], meshDirectory: File)(implicit rng: Random): (Option[DataCollection], Seq[Throwable]) = {
     val meshFileNames = meshDirectory.listFiles().toSeq.filter(fn => fn.getAbsolutePath.endsWith(".vtk") || fn.getAbsolutePath.endsWith(".stl"))
-    val (meshes, ioErrors) = DataUtils.partitionSuccAndFailedTries(for (meshFn <- meshFileNames) yield { MeshIO.readMesh(meshFn) })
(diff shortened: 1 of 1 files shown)


src/main/scala/scalismo/statisticalmodel/dataset/DataCollection.scala
```diff
@@ -128,7 +128,9 @@ object DataCollection {
    */
   def fromMeshDirectory(referenceMesh: TriangleMesh[_3D], meshDirectory: File)(implicit rng: Random): (Option[DataCollection], Seq[Throwable]) = {
     val meshFileNames = meshDirectory.listFiles().toSeq.filter(fn => fn.getAbsolutePath.endsWith(".vtk") || fn.getAbsolutePath.endsWith(".stl"))
-    val (meshes, ioErrors) = DataUtils.partitionSuccAndFailedTries(for (meshFn <- meshFileNames) yield { MeshIO.readMesh(meshFn) })
+    val (meshes, ioErrors) = DataUtils.partitionSuccAndFailedTries(for (meshFn <- meshFileNames) yield {
+      MeshIO.readMesh(meshFn).map(m => TriangleMesh3D(m.pointSet, referenceMesh.triangulation))
```
Andreas-Forster: I think it might be a good idea to add here a check if the triangulations are identical. Otherwise, this can introduce weird bugs if the user loads data which is not registered.

thogerig: Thanks for the comment. Already now it has not been possible to use "unregistered" meshes (the check is if they have the same amount of points), because in the method `fromMeshSequence()` this is checked. This fix, therefore, would not introduce an additional bug.

Andreas-Forster: I agree with your last comment.

PR #268, opened 2019-05-14
added explicit jvm-target directive for Scala 2.12 and higher

marcelluethi:
The target jvm was not set explicitly set for Scala 2.12 or higher. This results in runtime errors for users of Java 8, when a new java version is used to compile and publish Scalismo. 

This PR explicitly sets the target and thus avoids such errors. 

As the currently published version 0.17.1 was published using Java 8, no action is required on the publishing side.


marcelluethi:
@Ghazi-Bouabene could you have a look at this PR and merge it if you are okay with it?

PR #273, opened 2019-07-05
corrected computation of continuous indices in image domains. fixes #272

### LINKED ISSUE #272: findClosestPoint on images with negative direction matrix
The findClosestPoint function on an image domain does not work correctly on images, whose direction matrix is negative. The reason for this is that the internal function ```pointToContinuousIndex``` does not take the direction matrix into account when computing the continuous indices.

marcelluethi:
Due to a bug in the internal method ```pointToContinuousIndex```, finding closest points for images with a negative direction matrix did not work. The reason is that this method did not properly take the anisotropic transformation into account, which scalismo uses for computing the place and orientation of the domain. 

In the context of this bugfix, I also fixed an issue with the signature of the inverse of the ```AnisotropicSimilarityTransformation```, which returned the generic type ```Transformation``` rather than the specific type ```AnisotropicSimilarityTransformation```.

### DIFF
--- a/src/main/scala/scalismo/registration/TransformationSpace.scala
+++ b/src/main/scala/scalismo/registration/TransformationSpace.scala
@@ -796,18 +796,20 @@ case class AnisotropicScalingSpace[D: NDSpace]() extends TransformationSpace[D] 
  * The order of the rigid transform in this case is also fixed : first rotate then translate.
  *
  */
-trait AnisotropicSimilarityTransformation[D] extends CompositeTransformation[D] with CanInvert[D]
+trait AnisotropicSimilarityTransformation[D] extends CompositeTransformation[D] with CanInvert[D] {
+  override def inverse : AnisotropicSimilarityTransformation[D]
+}
 
 private class RigidTransformationThenAnisotropicScaling[D: NDSpace](anisotropicScaling: AnisotropicScalingTransformation[D], rigidTransform: RigidTransformation[D])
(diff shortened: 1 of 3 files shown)


src/main/scala/scalismo/registration/TransformationSpace.scala
```diff
@@ -796,18 +796,20 @@ case class AnisotropicScalingSpace[D: NDSpace]() extends TransformationSpace[D]
...
-trait AnisotropicSimilarityTransformation[D] extends CompositeTransformation[D] with CanInvert[D]
+trait AnisotropicSimilarityTransformation[D] extends CompositeTransformation[D] with CanInvert[D] {
+  override def inverse : AnisotropicSimilarityTransformation[D]
+}
 
 private class RigidTransformationThenAnisotropicScaling[D: NDSpace](anisotropicScaling: AnisotropicScalingTransformation[D], rigidTransform: RigidTransformation[D])
     extends CompositeTransformation[D](anisotropicScaling, rigidTransform) with AnisotropicSimilarityTransformation[D] {
 
-  def inverse: AnisotropicScalingThenRigidTransformation[D] = new AnisotropicScalingThenRigidTransformation[D](rigidTransform.inverse, anisotropicScaling.inverse)
+  override def inverse: AnisotropicScalingThenRigidTransformation[D] = new AnisotropicScalingThenRigidTransformation[D](rigidTransform.inverse, anisotropicScaling.inverse)
```
DanaRahbani: This might be dangerous, as we do not restrict AnisotropicScalingTransformations to be strictly positive. But let's discuss that in another issue.

PR #275, opened 2019-07-10
Fix neighbor query with VantagePointTree: fixes #274

### LINKED ISSUE #274: VantagePointTree.findEpsilonNeighbours fails when the point is in the tree
The next test because when the objective point is already in the tree.If I put a point not in the tree works ok
```
it should "find the points in a ball around excluding himself" in {

    Given("some points")
    val points = Seq(
      RNDensePoint(0, 0),
      RNDensePoint(0, 1),
      RNDensePoint(0, 2)
    )

    And("the tree with the points")
    val tree = VantagePointTree(points, RNDensePoint.euclidean)

    And("an objetive point")
    val objetive = RNDensePoint(0, 1)

    And("the ball radio")
    val radio = 4.0

    And("the expected results")
    val expected = Set(
      RNDensePoint(0, 0),
      RNDensePoint(0, 1),
      RNDensePoint(0, 2)
    )

    When("the tree find points in the ball")
    val closestPoints = tree.findEpsilonNeighbours(objetive, radio)

    Then("the nearest must be the expected")
    closestPoints.toSet shouldBe expected
  }
```

sschoenborn:
The VantagePointTree can be used to find neighboring points efficiently. However, the query did not work properly, see issue #274 

The query did not retrieve multiple, different points at the same distance to the query point. The reason for this is an improper use of a partial ordering with `SortedSet`. `SortedSet` assumes a total order and considers elements which compare equally to be equal and will only keep one of them in the set. When used with distance, different points at the same distance will be considered equal and not be added to the result set. 

Suggested fix: use a `PriorityQueue` which can keep multiple elements with the same priority (=distance in VP). I did not do any performance evaluations, but I assume the impact to be small.

Fixes #274


marcelluethi:
@sschoenborn Thanks for the bugfix and detailed description. I do not feel competent to review it and let Andi do this. But I changed the target branch to release-0.17. I plan to do a bugfix release within the next few days, where this could be a part of.

marcelluethi:
This PR was manually merged.

PR #276, opened 2019-07-17
Methods for decimating triangle meshes and statistical mesh models

marcelluethi:
Being able to painlessly reducing the size of meshes and models is of great importance when implementing efficient sampling-based pipelines. This PR proposes to add methods to decimate meshes and statistical mesh models.

As a helper function, a new interpolator, ```TriangleMeshInterpolator```, which interpolates on
the surface defined on a triangle mesh is implemented.

### DIFF
--- a/src/main/scala/scalismo/common/interpolation/TriangleMeshInterpolator.scala
+++ b/src/main/scala/scalismo/common/interpolation/TriangleMeshInterpolator.scala
@@ -1,0 +1,42 @@
+package scalismo.common.interpolation
+
+import scalismo.common.{DiscreteField, Field, RealSpace, UnstructuredPointsDomain}
+import scalismo.geometry.{Point, _3D}
+import scalismo.mesh.{SurfacePointProperty, TriangleMesh}
+import scalismo.mesh.boundingSpheres.{ClosestPointInTriangle, ClosestPointIsVertex, ClosestPointOnLine}
+import scalismo.numerics.ValueInterpolator
+
+/**
+  * Interpolates a given discrete field defined on the vertices
+  * of a triangle mesh (i.e. a MeshField) by means of a surface interpolation
+  * on the surface defined by the mesh.
+  *
+  * @param mesh The mesh on which the interpolation is performed.
+  */
+case class TriangleMeshInterpolator[A : ValueInterpolator](mesh : TriangleMesh[_3D]) extends
+  FieldInterpolator[_3D, UnstructuredPointsDomain[_3D], A] {
+
+  override def interpolate(field: DiscreteField[_3D,
+    UnstructuredPointsDomain[_3D], A]): Field[_3D, A] = {
+
+    // for this method to make sense, the field needs to be defined
+    // on the mesh. There is no good way to check this rigorously.
+    // A simple sanity check is, however, that the number of points
+    // of the domain is the same.
+    require(mesh.pointSet.numberOfPoints == field.values.size)
+
+    val fieldOnSurface =
+      SurfacePointProperty(mesh.triangulation, field.values.toIndexedSeq)
+
+    val f = (pt: Point[_3D]) =>
+      mesh.operations.closestPointOnSurface(pt) match {
+        case ClosestPointIsVertex(pt, _, id) => fieldOnSurface(id)
+        case ClosestPointInTriangle(pt, _, id, bc) =>
+          fieldOnSurface(id, bc)
+        case ClosestPointOnLine(pt, _, (id0, id1), d) =>
+          ValueInterpolator[A].blend(fieldOnSurface(id0), fieldOnSurface(id1), d)
+      }
+    Field(RealSpace[_3D], f)
+  }
+}
+
--- a/src/main/scala/scalismo/common/interpolation/TriangleMeshInterpolator.scala
+++ b/src/main/scala/scalismo/common/interpolation/TriangleMeshInterpolator.scala
@@ -1,23 +1,21 @@
 package scalismo.common.interpolation
 
-import scalismo.common.{DiscreteField, Field, RealSpace, UnstructuredPointsDomain}
-import scalismo.geometry.{Point, _3D}
-import scalismo.mesh.{SurfacePointProperty, TriangleMesh}
-import scalismo.mesh.boundingSpheres.{ClosestPointInTriangle, ClosestPointIsVertex, ClosestPointOnLine}
+import scalismo.common.{ DiscreteField, Field, RealSpace, UnstructuredPointsDomain }
+import scalismo.geometry.{ Point, _3D }
+import scalismo.mesh.{ SurfacePointProperty, TriangleMesh }
+import scalismo.mesh.boundingSpheres.{ ClosestPointInTriangle, ClosestPointIsVertex, ClosestPointOnLine }
 import scalismo.numerics.ValueInterpolator
 
 /**
-  * Interpolates a given discrete field defined on the vertices
-  * of a triangle mesh (i.e. a MeshField) by means of a surface interpolation
-  * on the surface defined by the mesh.
-  *
-  * @param mesh The mesh on which the interpolation is performed.
-  */
-case class TriangleMeshInterpolator[A : ValueInterpolator](mesh : TriangleMesh[_3D]) extends
-  FieldInterpolator[_3D, UnstructuredPointsDomain[_3D], A] {
+ * Interpolates a given discrete field defined on the vertices
+ * of a triangle mesh (i.e. a MeshField) by means of a surface interpolation
+ * on the surface defined by the mesh.
+ *
+ * @param mesh The mesh on which the interpolation is performed.
+ */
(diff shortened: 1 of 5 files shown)


src/main/scala/scalismo/common/interpolation/TriangleMeshInterpolator.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+case class TriangleMeshInterpolator[A: ValueInterpolator](mesh: TriangleMesh[_3D]) extends FieldInterpolator[_3D, UnstructuredPointsDomain[_3D], A] {
+
+  override def interpolate(field: DiscreteField[_3D, UnstructuredPointsDomain[_3D], A]): Field[_3D, A] = {
+
+    // for this method to make sense, the field needs to be defined
+    // on the mesh. There is no good way to check this rigorously.
+    // A simple sanity check is, however, that the number of points
+    // of the domain is the same.
+    require(mesh.pointSet.numberOfPoints == field.values.size)
```
Andreas-Forster: For me, a better test would check that the two domains are equal. Just the number of points seems to be a bit a weak test. At least in our lab I agree, even the weaker test might catch near all errors we could make easily.

src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -178,5 +179,27 @@ class TriangleMesh3DOperations(private val mesh: TriangleMesh3D) {
...
+  /**
+   * Attempts to reduce the number of vertices of a mesh to the given number of vertices.
+   *
+   * @param targetedNumberOfVertices The targeted number of vertices. Note that it is not guaranteed
+   *                                 that this number is reached exactly
+   * @return The decimated mesh
+   */
+  def decimate(targetedNumberOfVertices: Int): TriangleMesh[_3D] = {
+    val refVtk = MeshConversion.meshToVtkPolyData(mesh)
+    val decimatePro = new vtk.vtkDecimatePro()
```
Andreas-Forster: I am not qualified to judge the different mesh reductions implemented in vtk. So I rely on your choice here.

src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -181,24 +181,24 @@ class TriangleMesh3DOperations(private val mesh: TriangleMesh3D) {
...
+
+    val reductionRate = 1.0 - (targetedNumberOfVertices / mesh.pointSet.numberOfPoints)
```
DanaRahbani: I just tried this out. I had to change the last value to: mesh.pointSet.numberOfPoints.toDouble. Otherwise that fraction is always 0 and reductionRate is always 1.0

marcelluethi: @DanaRahbani Thanks for testing and pointing this out. This was really a beginner's mistake 😟  The bug should now be fixed and the corresponding test improved.

madsendennis:
decimate_face_10000_left
decimate_face_10000_right
decimate_faceFull

I just tried the decimating method on the bfm face12. Reduction from initial ~28000 points to target 10000. As such the decimation works well - but the decimated reference mesh does not look good. I am not familiar with the decimation methods in vtk, but hopefully, there is a better method which would make more "smooth" results. On the large smooth areas on the chins, the reduction seems very random. It is furthermore not consistent from left to right side.

madsendennis:
decimate_face_10000_QuadricDecimation
The "quadratic edge collapse decimation" available in MeshLab seems to me to be doing a better job. It also allows to set a target reduction as in the current implementation. 
Also available in vtk: https://vtk.org/doc/release/5.2/html/a01154.html

marcelluethi:
Thanks for testing this and your feedback. You are right, the decimated meshes are not visually pleasing and Quadricedgecollapse usually leads to results that look better. 

Before we swap it out, it would be good to check which filter leads to the more accurate surface approximation, as this is our main concern. This could be tested by measuring the surface distance between the approximated and the original mesh after decimation. As you have all the program ready, could I ask you to quickly try this out?

madsendennis:
Results for face reference decimation. Provided is both the distance from the decimated mesh to the original reference and from the reference to the decimated mesh. 
In some regards, the current implementation is better as it has an average distance of 0 from the decimated mesh to the original. But going the other way around it also shows that it lose a lot of details - when approximating to 1300 points it almost has an error of 1 mm on average.

MeshLab refers to using QECD in MeshLab and Scalismo refers to using the current implemented DecimatePro in Scalismo. 

Points in reference: 53149
Target points: 1300
Avg dist, MeshLab decimated (1300 points) to ref: 0.09334155237642557 mm
Avg dist, ref to MeshLab decimated (1300 points): 0.06025832331278917 mm
Avg dist, Scalismo decimated (1380 points) to ref: 0.0 mm
Avg dist, ref to Scalismo decimated (1380 points): 0.7091218916645096 mm
Target points: 2500
Avg dist, MeshLab decimated (2583 points) to ref: 0.040145941921963905 mm
Avg dist, ref to MeshLab decimated (2583 points): 0.030154863162202318 mm
Avg dist, Scalismo decimated (2602 points) to ref: 0.0 mm
Avg dist, ref to Scalismo decimated (2602 points): 0.15154744134865186 mm
Target points: 5000
Avg dist, MeshLab decimated (5123 points) to ref: 0.015037470600761865 mm
Avg dist, ref to MeshLab decimated (5123 points): 0.018201433639048956 mm
Avg dist, Scalismo decimated (5118 points) to ref: 0.0 mm
Avg dist, ref to Scalismo decimated (5118 points): 0.05806246788153554 mm
Target points: 10000
Avg dist, MeshLab decimated (10161 points) to ref: 0.006023943304438063 mm
Avg dist, ref to MeshLab decimated (10161 points): 0.006247378103767976 mm
Avg dist, Scalismo decimated (10123 points) to ref: 0.0 mm
Avg dist, ref to Scalismo decimated (10123 points): 0.02430256871826957 mm

marcelluethi:
Thanks a lot, Dennis.
Given that the error for QuadricEdgeCollapse are so small, it would make sense to switch. We should, however, also check directly with the vtk implementation. I am not sure if it is as accurate as the meshlab one.

BernhardEgger:
QuadradicEdgeCollapse in Meshlab has tons of settings and some of them are very useful in practice. It is e.g. nice if there are no new points introduced. I don't know if those settings are available in VTK but if they are it would probably be nice to give the user control over those.

Going one step further would be to allow to reduce models. This is definitely not the aim of this PR but might be worth considering later. A nice paper that I just found 2 weeks ago that I can't believe I didn't find earlier:
https://doi.org/10.1109/ICCV.2011.6126252
@inproceedings{patel2011simplification,
  title={Simplification of 3D morphable models},
  author={Patel, Ankur and Smith, William AP},
  booktitle={2011 International Conference on Computer Vision},
  pages={271--278},
  year={2011},
  organization={IEEE}
}

madsendennis:
I tried swapping out vtkDecimatePro with vtkQuadricDecimation in the mesh.operation.decimate function. The result is very similar to what was obtained in MeshLab. Obviously, the goal for each decimation method is different, and might both be useful for different applications. In general, I though think that QuadricDecimation is more useful. 

Original number of vertices: 53149
DecPRO
Avg dist, decimated (1380 points) to ref: 0.0 mm
Avg dist, ref to decimated (1380 points): 0.7091218916645096 mm
DecQuad
Avg dist, decimated (1362 points) to ref: 0.07718615301348894 mm
Avg dist, ref to decimated (1362 points): 0.06320852335916823 mm
DecPRO
Avg dist, decimated (2602 points) to ref: 0.0 mm
Avg dist, ref to decimated (2602 points): 0.15154744134865186 mm
DecQuad
Avg dist, decimated (2587 points) to ref: 0.036011943397666515 mm
Avg dist, ref to decimated (2587 points): 0.031976004107904465 mm
DecPRO
Avg dist, decimated (5118 points) to ref: 0.0 mm
Avg dist, ref to decimated (5118 points): 0.05806246788153554 mm
DecQuad
Avg dist, decimated (5105 points) to ref: 0.015932210818674596 mm
Avg dist, ref to decimated (5105 points): 0.015176981666402101 mm
DecPRO
Avg dist, decimated (10123 points) to ref: 0.0 mm
Avg dist, ref to decimated (10123 points): 0.02430256871826957 mm
DecQuad
Avg dist, decimated (10108 points) to ref: 0.0057847374768059175 mm
Avg dist, ref to decimated (10108 points): 0.006329213753073516 mm

I do not know how user friendly it would be to provide handles to how the different parts are weighted in the error metric. The error metric as based on the following properties which can turned on (default) or off as well as individually scaled: 
- SetScalarsWeight
- SetVectorsWeight
- SetNormalsWeight
- SetTCoordsWeight
- SetTensorsWeight

marcelluethi:
Thanks, Dennis. I agree, we should switch to Quadric edge. Can you do the PR?
I personally would not expose the weight options in an initial version. They seem not to be very intuitive (I at least would not know how to use them). Furthermore, it is always easy to add them later if we really see the need.

PR #278, opened 2019-07-20
Implemented functionality to read and write deformation models.

marcelluethi:
The implementation is done in terms of a more generic method, which can write
any  GP defined over an image domain as a statismo file.

### DIFF
--- a/src/main/scala/scalismo/io/StatisticalModelIO.scala
+++ b/src/main/scala/scalismo/io/StatisticalModelIO.scala
@@ -18,46 +18,99 @@ package scalismo.io
 import java.io._
 import java.util.Calendar
 
-import breeze.linalg.{ DenseMatrix, DenseVector }
+import breeze.linalg.{DenseMatrix, DenseVector}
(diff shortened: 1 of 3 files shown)


src/main/scala/scalismo/io/StatisticalModelIO.scala
```diff
@@ -342,4 +387,231 @@ object StatismoIO {
...
+        case (0, 9) => Success(pcaBasisMatrix)
+        case (0, 8) => Success(extractOrthonormalPCABasisMatrix(pcaBasisMatrix, pcaVarianceVector)) // an old statismo version
+        case v => Failure(new Throwable(s"Unsupported version ${v._1}.${v._2}"))
+      }
+
+      _ <- Try {
+        h5file.close()
+      }
+    } yield {
+      // statismo stores the mean as the point position, not as a displacement on the reference.
```
Andreas-Forster: This comment seems to be a left over. Can't this be removed? Or should it be accounted for that in the following lines?

PR #285, opened 2019-08-08
Procrustes mesh distance fix.

madsendennis:
Updated from using average distance to surface to using the distance between corresponding points. Fix for issue #196.

### DIFF
--- a/src/main/scala/scalismo/mesh/MeshMetrics.scala
+++ b/src/main/scala/scalismo/mesh/MeshMetrics.scala
@@ -16,21 +16,21 @@
 package scalismo.mesh
 
 import scalismo.common.BoxDomain
-import scalismo.geometry.{ Point, _3D }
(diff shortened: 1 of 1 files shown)


marcelluethi:
Thanks Dennis for fixing this. In principle this looks good.

Behzad recently brought to my attention the difference between partial procrustes alignment and full procrustes alignment. In the former, scale is held fixed in the transformation, whereas in the latter the scale would be optimized over when computing the optimal alignment. So this method would correspond to partial Procrustes distance. 

In Scalismo we are consistent in using the partial procrustes method everywhere, so I don't think we should change the name of all the methods. But we should at least add a comment and be aware of this distinction.

marcelluethi:
BTW, the PR fails on travis due to formatting issues. To force formatting by Scalariform you could run the unit tests or call
```
sbt compile:scalariformFormat
sbt test:scalariformFormat
```

PR #287, opened 2019-08-14
Parameter D in Eigenbasis Calculation from Pivoted Cholesky Output

### LINKED ISSUE #164: Parameters of Pivoted Cholesky undocumented and/or unintuitive
The function
```
PivotedCholesky.computeApproximateEigGeneric
```
takes a parameter ```D```. It is from the outside not clear what this parameter does and why 
it is needed. Either it has to be documented, or (if possible) hidden from the interface.

thogerig:
Fixes #164 

The mass matrix parameter D is in our situation a scalar, which is set to 1.0. Since there is no occasion to be found where we would have to change this, I would also propose to remove the exposure of this parameter. 

In a future case, in which the current calculation needs to be adapted with a different D, I would propose to define it as a separate function.

### DIFF
--- a/src/main/scala/scalismo/numerics/PivotedCholesky.scala
+++ b/src/main/scala/scalismo/numerics/PivotedCholesky.scala
@@ -183,12 +183,12 @@ object PivotedCholesky {
     computeApproximateCholeskyGeneric[Int](kernel, indices, stoppingCriterion)
   }
 
-  private def computeApproximateEigGeneric[A](k: (A, A) => Double, xs: IndexedSeq[A], D: Double, sc: StoppingCriterion) = {
+  private def computeApproximateEigGeneric[A](k: (A, A) => Double, xs: IndexedSeq[A], sc: StoppingCriterion) = {
 
     val PivotedCholesky(l, _, _) = computeApproximateCholeskyGeneric(k, xs, sc)
 
-    val LD = l(::, 0 until l.cols).t *:* D
-    val phi: DenseMatrix[Double] = LD * l(::, 0 until l.cols)
+    val Lt = l(::, 0 until l.cols).t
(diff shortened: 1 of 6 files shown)


marcelluethi:
@thogerig Thanks a lot for addressing this issue. 

I agree that this parameter should not be an argument that the user can chose. 
Yet, as it seems to be important in the calculation and seems to have a meaning, I would prefer not to drop it altogether. 
How about introducing it as a value ```val D = 1``` inside the function? We would then be reminded whenever we look at the method, that there is such a parameter ```D``` which can play a role. Ideally, we would also add a comment explaining its effect and why we keep it fixed.

thogerig:
@marcelluethi Thanks for your input.

To me this seems like a good idea. I have updated the pull request with a reintroduction of `D` and a description with a pointer to further literature.

src/main/scala/scalismo/numerics/PivotedCholesky.scala
```diff
@@ -183,12 +183,19 @@ object PivotedCholesky {
...
 
     val PivotedCholesky(l, _, _) = computeApproximateCholeskyGeneric(k, xs, sc)
 
-    val LD = l(::, 0 until l.cols).t *:* D
-    val phi: DenseMatrix[Double] = LD * l(::, 0 until l.cols)
+    // Parameter D is called mass matrix in the 'Galerkin discretization' of the eigenvalue problem.
+    // Here it can be written as a scalar value.
+    // For more information in regards to situations where D might be adapted, please start with
+    // "On the Low-rank Approximation by the PivotedCholesky Decomposition" (H. Harbrecht, M. Peters, R. Schneider)
+
```
marcelluethi: I find this comment somewhat confusing. Why is the matrix a scalar in this case? Where in the paper can we find the explanation? At least a first quick look at the paper does not reveal a discussion regarding the parameter D.

PR #293, opened 2019-11-13
TetrahedralMeshIO

Andreas-Forster:
This PR adds the functionality needed to read and write the class TetrahedralMesh.

Co-authored-by: Andreas-Forster <andreas.forster@unibas.ch>
Co-authored-by: Jean-Rassaire <jfrassaire@gmail.com>

### DIFF
--- a/src/main/scala/scalismo/utils/Conversions.scala
+++ b/src/main/scala/scalismo/utils/Conversions.scala
@@ -157,22 +157,15 @@ object VtkHelpers {
 
 object TetrahedronMeshConversion {
 
-  private def extractPointsAndCells(ug: vtkUnstructuredGrid, calculateNewDelaunayTetrahedrons: Boolean = false) = Try {
-    val newUg = if (calculateNewDelaunayTetrahedrons) {
-      val tetrahedralFilter = new vtkDelaunay3D
-      tetrahedralFilter.SetInputData(ug)
-      tetrahedralFilter.Update()
-      tetrahedralFilter.GetOutput()
-    } else ug
-
-    val grids = newUg.GetCells()
+  private def extractPointsAndCells(ug: vtkUnstructuredGrid) = Try {
+    val grids = ug.GetCells()
     val numGrids = grids.GetNumberOfCells()
 
-    val points = CommonConversions.vtkConvertPoints[_3D](newUg)
+    val points = CommonConversions.vtkConvertPoints[_3D](ug)
 
     val idList = new vtkIdList()
     val cells = for (i <- 0 until numGrids) yield {
-      newUg.GetCellPoints(i, idList)
+      ug.GetCellPoints(i, idList)
       if (idList.GetNumberOfIds() != 4) {
         throw new Exception("Not a tetrahedral mesh")
       }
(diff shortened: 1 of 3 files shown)


src/main/scala/scalismo/utils/Conversions.scala
```diff
@@ -153,6 +155,100 @@ object VtkHelpers {
 
 }
 
+object TetrahedronMeshConversion {
+
+  private def extractPointsAndCells(ug: vtkUnstructuredGrid, calculateNewDelaunayTetrahedrons: Boolean = false) = Try {
```
marcelluethi: I find the calculateNewDelauneyTetrahedrons here out of place. We are in the conversion method, which converts structures and should not change them. I would find it better to have this as part of the mesh operations.

src/main/scala/scalismo/utils/Conversions.scala
```diff
@@ -153,6 +155,100 @@ object VtkHelpers {
...
+                                            cells: IndexedSeq[TetrahedralCell]
+                                          ): (IndexedSeq[Point[_3D]], IndexedSeq[TetrahedralCell]) = {
+    val cellPointIds = cells.flatMap(_.pointIds).distinct
+    val oldId2newId = cellPointIds.zipWithIndex.map { case (id, index) => (id, PointId(index)) }.toMap
+    val newCells = cells.map(c => TetrahedralCell(oldId2newId(c.ptId1), oldId2newId(c.ptId2), oldId2newId(c.ptId3), oldId2newId(c.ptId4)))
+    val newPoints = cellPointIds.map(id => points(id.id))
+    (newPoints, newCells)
+  }
+
+  def vtkUnstructuredGridToTetrahedralMesh(ug: vtkUnstructuredGrid, recalculateTetrahedralization: Boolean = false): Try[TetrahedralMesh[_3D]] = {
```
marcelluethi: see my comment above. Do we really ever want to do a different tetrahedralization?

src/test/scala/scalismo/io/MeshIOTests.scala
```diff
@@ -154,4 +177,30 @@ class MeshIOTests extends ScalismoTestSuite {
...
+    val points = IndexedSeq.fill(N)(Point(
+      rng.scalaRandom.nextGaussian() * 2,
+      rng.scalaRandom.nextGaussian() * 1000,
+      rng.scalaRandom.nextGaussian() * 1000000
+    ))
+    val domain = UnstructuredPointsDomain(points)
+
+    // cells covering the complete cube
+    implicit def intToPointId(i: Int): PointId = PointId(i)
+    val T = 2
```
rassaire: why only 2 cells (T=2) from 200 points (N=200)?
For tetrahedral mesh, the number of cells is larger than the number of points

src/main/scala/scalismo/utils/Conversions.scala
Andreas-Forster: I agree that this should be moved or even dropped, at least for the moment. As it is not IO I will drop it from the PR for the moment.

src/main/scala/scalismo/utils/Conversions.scala
Andreas-Forster: I agree - see the answer above

src/test/scala/scalismo/io/MeshIOTests.scala
Andreas-Forster: In principle 2 is enough, while it actually is a change for a quick test that accidentally made it here. The mesh should not be anyhow reasonable, more the other way round, it should be special so that possible errors in the IO methods would show up quickly.

I set both to 200.

Andreas-Forster:
I agree with the reviewers' feedback. The last commit should have addressed all remarks made.

PR #295, opened 2019-11-14
add scalar volume mesh field IO and tests

DanaRahbani:
Co-authored-by: madsendennis <dennis.madsen@unibas.ch>

### DIFF
--- a/src/main/scala/scalismo/mesh/ScalarVolumeMeshField.scala
+++ b/src/main/scala/scalismo/mesh/ScalarVolumeMeshField.scala
@@ -20,6 +20,14 @@ import scalismo.geometry.{ Point, _3D }
 
 import scala.reflect.ClassTag
 
+/**
+  * 3-dimensional tetrahedral mesh with scalar values associated to mesh points.
+  *
+  * @tparam S type of the scalar values defined over the mesh (Short, Int, Float, Double)
+  * @constructor Returns a scalar volume mesh data given a tetrahedral mesh and an array of values.
+  * The number of values and mesh points must be equal.
+  */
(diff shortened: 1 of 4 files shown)


src/test/scala/scalismo/io/MeshIOTests.scala
```diff
@@ -204,4 +283,12 @@ class MeshIOTests extends ScalismoTestSuite {
     TetrahedralMesh3D(domain, list)
   }
 
+  def createRandomScalarVolumeMeshField(): ScalarVolumeMeshField[Int] = {
```
marcelluethi: this could be made private, as it is only used internally

src/main/scala/scalismo/mesh/ScalarVolumeMeshField.scala
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+package scalismo.mesh
+
+import scalismo.common.{ DiscreteScalarField, PointId, Scalar, ScalarArray, UnstructuredPointsDomain }
+import scalismo.geometry.{ Point, _3D }
+
+import scala.reflect.ClassTag
+
+case class ScalarVolumeMeshField[S: Scalar: ClassTag](mesh: TetrahedralMesh[_3D], override val data: ScalarArray[S])
```
marcelluethi: Here we should add a short description (api-doc) what the class does.

marcelluethi:
@DanaRahbani Can you run ```sbt scalariformFormat``` to fix the formatting and commit again? Then we are ready to merge.

PR #298, opened 2019-11-15
Barycentric interpolator

kahrpatrick:
Adds a Barycentric interpolator that can be used with tetrahedral meshes. Includes basic unit tests.

### DIFF
--- a/src/main/scala/scalismo/mesh/TetrahedralMesh.scala
+++ b/src/main/scala/scalismo/mesh/TetrahedralMesh.scala
@@ -134,6 +134,19 @@ case class TetrahedralMesh3D(pointSet: UnstructuredPointsDomain[_3D], tetrahedra
     math.abs(signedVolume)
   }
 
+  /** Returns the Barycentric coordinates of a point inside the indicated cell. */
+  def getBarycentricCoordinates(point: Point[_3D], tetrathedron: TetrahedralCell): Array[Double] = {
+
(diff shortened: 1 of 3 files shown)


src/main/scala/scalismo/mesh/TetrahedralMesh.scala
```diff
@@ -134,6 +134,19 @@ case class TetrahedralMesh3D(pointSet: UnstructuredPointsDomain[_3D], tetrahedra
...
+  /** Returns the Barycentric coordinates of a point inside the indicated cell. */
+  def getBarycentricCoordinates(point: Point[_3D], tetrathedron: TetrahedralCell): Array[Double] = {
+
+    val a = pointSet.point(tetrathedron.ptId1).toVector
+    val b = pointSet.point(tetrathedron.ptId2).toVector
+    val c = pointSet.point(tetrathedron.ptId3).toVector
+    val d = pointSet.point(tetrathedron.ptId4).toVector
+
+    val barycentricCoordinates = new Array[Double](4)
+    new vtkTetra().BarycentricCoords(point.toArray, a.toArray, b.toArray, c.toArray, d.toArray, barycentricCoordinates)
```
marcelluethi: For the moment it is okay to call vtk here. But we should replace it at some point with out own method, in order to gain some speedup. It is simple to implement. 

If we don't fix it now, we should at least add a !TODO marker here

PR #301, opened 2019-11-15
another attempt to clean up naming and some code duplications for svms

marcelluethi:


marcelluethi:
I will close this PR but merge its content on the console. The reason is that there seems to be some chaos with the develop branch, which needs to be taken care of.

PR #305, opened 2019-11-19
Sampling inside tetrahedral mesh

### LINKED ISSUE #299: Points sampled inside tetrahedron not always inside
I noticed that points returned by  `samplePointInTetrahedralCell` are sometimes outside of a cell.
The following test currently fails:

`implicit val rng = Random(42L)`

`it("points sampled inside a tetrahedron should not be outise of tetrahedron") {
      val tetrahedralMesh = createTetrahedronsInUnitCube()
      val cell = tetrahedralMesh.tetrahedralization.tetrahedron(TetrahedronId(0))
      val point =  tetrahedralMesh.samplePointInTetrahedralCell(cell)
      assert(tetrahedralMesh.isInsideTetrahedralCell(point, cell))
 }`

rassaire:
fixes #299

### DIFF
--- a/src/main/scala/scalismo/mesh/TetrahedralMesh.scala
+++ b/src/main/scala/scalismo/mesh/TetrahedralMesh.scala
@@ -175,7 +175,25 @@ case class TetrahedralMesh3D(pointSet: UnstructuredPointsDomain[_3D], tetrahedra
    * @param tc   Tetrahedral cell in which to draw a random point
    * @param rnd implicit Random object
    */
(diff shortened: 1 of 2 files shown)


rassaire:
I agree, "while loop" should be avoided. I have tested the proposed one and it works.

Andreas-Forster:
Checks failed due to mal formatted test sources. Commit follows in an instant. Missed to execute both:

- `scalariformFormat`
- `test:scalariformFormat`

PR #310, opened 2019-12-17
StatisticalVolumeIntensityModel 

kahrpatrick:
This is a first implementation of a StatisticalVolumeIntensityModel. Dot not merge this pull-request yet, as many elements are still missing.

A simple model can be created from a reference ScalarVolumeMeshField, a StatisiticalVolumeMeshModel (shape) and a DiscreteLowRankGaussianProcess (intensity).

A vectorizer for objects of type Scalar is also provided.

### DIFF
--- a/src/main/scala/scalismo/common/Scalar.scala
+++ b/src/main/scala/scalismo/common/Scalar.scala
@@ -435,16 +435,4 @@ object ScalarArray {
     }
   }
 
-  case class ScalarVectorizer[S: Scalar]() extends Vectorizer[S] {
-    override def dim: Int = 1
-
-    def toArray(v: S): Array[Double] = Array[Double](Scalar[S].toDouble(v))
-
-    override def vectorize(v: S): DenseVector[Double] = new DenseVector(toArray(v))
-
-    override def unvectorize(d: DenseVector[Double]): S = {
-      Scalar[S].fromDouble(d(0))
-    }
-  }
-
 }
--- a/src/main/scala/scalismo/common/Scalar.scala
+++ b/src/main/scala/scalismo/common/Scalar.scala
@@ -423,21 +423,15 @@ object ScalarArray {
     }
   }
 
-  class ScalarVectorizer[S: Scalar] extends Vectorizer[S] {
+  case class ScalarVectorizer[S: Scalar]() extends Vectorizer[S] {
     override def dim: Int = 1
 
     def toArray(v: S): Array[Double] = Array[Double](Scalar[S].toDouble(v))
+
     override def vectorize(v: S): DenseVector[Double] = new DenseVector(toArray(v))
 
     override def unvectorize(d: DenseVector[Double]): S = {
       Scalar[S].fromDouble(d(0))
-    }
-
-    override def equals(that: Any): Boolean = {
-      that match {
-        case t: ScalarVectorizer[S] => true
-        case _ => false
-      }
     }
   }
 
--- a/src/main/scala/scalismo/common/Scalar.scala
+++ b/src/main/scala/scalismo/common/Scalar.scala
@@ -15,6 +15,7 @@
  */
 package scalismo.common
 
+import breeze.linalg.DenseVector
 import scalismo.utils.ArrayUtils
 import spire.math._
 
@@ -419,6 +420,24 @@ object ScalarArray {
         v.convertArray[T](array, { t =>
           t
         })
+    }
+  }
+
+  class ScalarVectorizer[S: Scalar] extends Vectorizer[S] {
+    override def dim: Int = 1
+
+    def toArray(v: S): Array[Double] = Array[Double](Scalar[S].toDouble(v))
+    override def vectorize(v: S): DenseVector[Double] = new DenseVector(toArray(v))
+
+    override def unvectorize(d: DenseVector[Double]): S = {
+      Scalar[S].fromDouble(d(0))
+    }
+
+    override def equals(that: Any): Boolean = {
+      that match {
+        case t: ScalarVectorizer[S] => true
+        case _ => false
+      }
     }
   }
 
--- a/src/main/scala/scalismo/statisticalmodel/SVIMCoefficients.scala
+++ b/src/main/scala/scalismo/statisticalmodel/SVIMCoefficients.scala
@@ -12,16 +12,16 @@ case class SVIMCoefficients(shape: DenseVector[Double],
 
 object SVIMCoefficients {
   def apply(shape: IndexedSeq[Double],
-    color: IndexedSeq[Double]) = new SVIMCoefficients(DenseVector(shape.toArray), DenseVector(color.toArray))
+            intensity: IndexedSeq[Double]) = new SVIMCoefficients(DenseVector(shape.toArray), DenseVector(intensity.toArray))
 
   def apply(shape: DenseVector[Double],
-    color: DenseVector[Double]) = new SVIMCoefficients(shape, color)
+            intensity: DenseVector[Double]) = new SVIMCoefficients(shape, intensity)
 
   /** get 0 coefficients of specified length */
   def zeros(shapeComponents: Int,
-    colorComponents: Int): SVIMCoefficients = {
+            intensityComponents: Int): SVIMCoefficients = {
     new SVIMCoefficients(
       DenseVector.zeros(shapeComponents),
-      DenseVector.zeros(colorComponents))
+      DenseVector.zeros(intensityComponents))
   }
 }
--- a/src/main/scala/scalismo/statisticalmodel/SVIMCoefficients.scala
+++ b/src/main/scala/scalismo/statisticalmodel/SVIMCoefficients.scala
@@ -1,0 +1,27 @@
+package scalismo.statisticalmodel
+
+import breeze.linalg.DenseVector
+
+case class SVIMCoefficients(shape: DenseVector[Double],
+    intensity: DenseVector[Double]) {
+  def *(f: Float): SVIMCoefficients = this * f.toDouble
+  def *(d: Double): SVIMCoefficients = copy(shape = shape * d, intensity = intensity * d)
+  def +(other: SVIMCoefficients): SVIMCoefficients = copy(shape = shape + other.shape, intensity = intensity + other.intensity)
+  def -(other: SVIMCoefficients): SVIMCoefficients = copy(shape = shape - other.shape, intensity = intensity - other.intensity)
+}
+
+object SVIMCoefficients {
+  def apply(shape: IndexedSeq[Double],
+    color: IndexedSeq[Double]) = new SVIMCoefficients(DenseVector(shape.toArray), DenseVector(color.toArray))
+
+  def apply(shape: DenseVector[Double],
+    color: DenseVector[Double]) = new SVIMCoefficients(shape, color)
+
+  /** get 0 coefficients of specified length */
+  def zeros(shapeComponents: Int,
+    colorComponents: Int): SVIMCoefficients = {
+    new SVIMCoefficients(
+      DenseVector.zeros(shapeComponents),
+      DenseVector.zeros(colorComponents))
+  }
+}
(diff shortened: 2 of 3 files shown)


rassaire:
Thanks, Patrick for the pull request.  Actually the StatisticalVolumeIntensityModel is combined of two separate models (shape, intensity). Its does not include the correlation between shape and intensity which is good because we can fix the parameter of one model and get the other model.  But I thought we could also have a model of shape and intensity with correlation embedded.  
That is representing a feature as a dense vector of shape plus intensity. That what is  I have implemented, in this case, deformation fields are defined from the 3D domain to R^4.  I also defined a transformation for a scalarVolumeMeshField, that transforms the domain as well as the intensity.

hnbonsou:
Thank Patrick,
I have the same concern as Jean mentioned, embed the shape and intensity in the same model working on 4D, or have a vectorizer that manages (domain, intensity) object.

marcelluethi:
@hnbonsou @rassaire I understand your use case and think that such combined models are very useful and needed. Yet as an implementation in Scalismo, I think it makes sense to start with the two independent model. It seems to easier and more general to implement. The combined use case can easily be implemented (even from a user application) once the separate models are there. Also the visualization from ScalismoUI can easily be coupled from the user code, by simply projecting into the common model. 

The converse is currently not true. With the current design of GPs in Scalismo, it is not clear how to separate a combined model  into intensity and shape components. Also, it would not be clear how to integrate the combined model into the current design of the ui. 

My suggestion is to start experimenting with this approach, keep the 4D model at the beginning in user code (and not core scalismo) and once we have sufficient experience and see the need for more direct support, we discuss the integration in scalismo.

rassaire:
Okay Marcel fair enough, as i said the combined of separate models is still useful.

src/main/scala/scalismo/common/Scalar.scala
```diff
@@ -408,4 +409,22 @@ object ScalarArray {
...
+    override def dim: Int = 1
+
+    def toArray(v: S): Array[Double] = Array[Double](Scalar[S].toDouble(v))
+    override def vectorize(v: S): DenseVector[Double] = new DenseVector(toArray(v))
+
+    override def unvectorize(d: DenseVector[Double]): S = {
+      Scalar[S].fromDouble(d(0))
+    }
+
+    override def equals(that: Any): Boolean = {
```
marcelluethi: whenever equals is implemented, hashcode should also be implemented. 
But easier still would be to make it a case class

src/main/scala/scalismo/statisticalmodel/StatisticalVolumeIntensityModel.scala
```diff
@@ -0,0 +1,100 @@
...
+import breeze.linalg.DenseVector
+import scalismo.common._
+import scalismo.geometry._
+import scalismo.mesh._
+import scalismo.utils.Random
+
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe.TypeTag
+
+trait StatisticalVolumeIntensityModel[S] {
```
marcelluethi: What is the reason that it is split into a trait and an implementation class? Do we expect other subclasses? Would it even be possible to have other subclasses, which satisfy these type signatures?

src/main/scala/scalismo/statisticalmodel/SVIMCoefficients.scala
```diff
@@ -0,0 +1,27 @@
...
+  def +(other: SVIMCoefficients): SVIMCoefficients = copy(shape = shape + other.shape, intensity = intensity + other.intensity)
+  def -(other: SVIMCoefficients): SVIMCoefficients = copy(shape = shape - other.shape, intensity = intensity - other.intensity)
+}
+
+object SVIMCoefficients {
+  def apply(shape: IndexedSeq[Double],
+    color: IndexedSeq[Double]) = new SVIMCoefficients(DenseVector(shape.toArray), DenseVector(color.toArray))
+
+  def apply(shape: DenseVector[Double],
+    color: DenseVector[Double]) = new SVIMCoefficients(shape, color)
```
rassaire: to be consistent, replace "color" with "intensity"

src/main/scala/scalismo/statisticalmodel/StatisticalVolumeIntensityModel.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  def referenceMesh: TetrahedralMesh3D
+
+  def referenceMeshField: ScalarVolumeMeshField[S]
+
+  def shape: StatisticalVolumeMeshModel
+
+  def intensity: DiscreteLowRankGaussianProcess[_3D, UnstructuredPointsDomain[_3D], S]
+
+  def landmarks: Map[String, Landmark[_3D]] = Map.empty[String, Landmark[_3D]]
```
marcelluethi: I don't think we need landmarks in the model. This makes sense for faces (in the MoMO implementation), as we are always modelling faces. But for most shapes we could not define a canonical set of (named) landmarks. And if we can, we can always store them in a separate file.

src/main/scala/scalismo/statisticalmodel/StatisticalVolumeIntensityModel.scala
```diff
@@ -0,0 +1,100 @@
...
+  def landmarksWithPointIds: Map[String, Option[PointId]] = landmarks.map { case (id, lm) => id -> referenceMesh.pointSet.pointId(lm.point) }
+
+  def landmarksWithClosestPointIds: Map[String, PointId] = landmarks.map { case (id, lm) => id -> referenceMesh.pointSet.findClosestPoint(lm.point).id }
+
+  def zeroCoefficients: SVIMCoefficients
+}
+
+object StatisticalVolumeIntensityModel {
+
+  def apply[S: Scalar: TypeTag: ClassTag](referenceMeshField: ScalarVolumeMeshField[S],
```
marcelluethi: It should be sufficient for the reference to be a TetrahedralMesh as it represents not an example function, but only the domain of the function.

src/main/scala/scalismo/statisticalmodel/StatisticalVolumeIntensityModel.scala
rassaire: what is the landmarks for?

src/main/scala/scalismo/statisticalmodel/StatisticalVolumeIntensityModel.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  def landmarks: Map[String, Landmark[_3D]] = Map.empty[String, Landmark[_3D]]
+
+  def mean: ScalarVolumeMeshField[S]
+
+  def instance(coefficients: SVIMCoefficients): ScalarVolumeMeshField[S]
+
+  def sample()(implicit rnd: Random): ScalarVolumeMeshField[S]
+
+  def withLandmarks(landmarksMap: Map[String, Landmark[_3D]]): StatisticalVolumeIntensityModel[S]
```
rassaire: same comment  above

src/main/scala/scalismo/statisticalmodel/SVIMCoefficients.scala
kahrpatrick: I agree, this is still a left-over from the MoMo class I took as a template.

src/main/scala/scalismo/statisticalmodel/StatisticalVolumeIntensityModel.scala
kahrpatrick: This makes sense. I will remove them.

kahrpatrick: Landmarks will be removed. See the comment by @marcelluethi above.

src/main/scala/scalismo/statisticalmodel/StatisticalVolumeIntensityModel.scala
rassaire: Doesn't the intensity model need a reference intensity,  even though it does not represent the domain? I though the reference intensities were subtracted from the targets ones i.e the deformation function values  are the differences between reference intensities and the target ones. In that case the reference intensity is needed as well (the reference should be a ScalarVolumeMeshField)

src/main/scala/scalismo/common/Scalar.scala
kahrpatrick: I did not think about the possibility of making it a case class. My implementation is consistent with `scalismo.geometry.EuclideanVector.VectorVectorizer`.

src/main/scala/scalismo/statisticalmodel/StatisticalVolumeIntensityModel.scala
marcelluethi: It is important to distinguish between the reference and the mean. It needs a mean intensity, but as this is provided separately.

rassaire: I though as for shape deformation function where values are differences between reference points and target points which give vectors.

marcelluethi: In the case of the deformation model, the mean is indeed a difference from the reference shape to the mean shape. In case of the intensities, it could also be implemented like this. I am not sure, however, if this is not more complicated than just providing the mean intensity directly. 

What do you think?

rassaire: I think for consistency, we do the same as in shape model. I don't think it will be complicated.

marcelluethi: Lets consider as a concrete example that we specify a zero-mean Gaussian process. If we use the scalars of the reference mesh, this would mean that our assumptions are on average the same as the ones on the reference. If we only have a mesh on the reference, our model would say that the intensities would in average be zero. 

Which one is more intuitive?
I tend to agree with you that using a ScalarVolumeMesh as a reference, and only modelling the difference is better. What do others think?

kahrpatrick: Considering your example, @rassaire's proposal makes sense. In that case the intensity model would be defined over vectors and not scalars?

marcelluethi: No, it can remain 1 dimensional. The difference here is just a scalar value (the difference to the mean intensity)

kahrpatrick: You mean the difference to the reference?

kahrpatrick: I updated the pull-request. The reference is again a  ScalarVolumeMeshField and intensities are generated by warping the reference intensities with the sampled differences.

src/main/scala/scalismo/common/Scalar.scala
kahrpatrick: I made it a case class now. We should not forget to apply the same change to `scalismo.geometry.EuclideanVector.VectorVectorizer`.

src/main/scala/scalismo/statisticalmodel/StatisticalVolumeIntensityModel.scala
hnbonsou: I agree using the reference image as a reference, the sampled intensity from such a model will be close to the original intensity range of the imaging modality, which is not always the case for the mean due to the correspondence accuracy.

marcelluethi:
There seems to be no objections with this PR. 

In order not to block progress, I propose to merge this PR and, together with the tetrahedralmesh operations and new sbt/Scalariform update,  use it as a basis for the milestone release M2. 

@kahrpatrick When you merge, can you squash all commits into 1 and write a good commit message summarizing your changes?

PR #315, opened 2020-01-16
removed dash from travis.yml, which was accidentally added

marcelluethi:


marcelluethi:
@kahrpatrick Thanks for pointing out the error in the travis file. It should now work as expected.

kahrpatrick:
Thanks for fixing the error. The reason why the build has failed is a remaining bug in my pull-request that I merged yesterday. I will commit a fix.

PR #320, opened 2020-01-27
Adds BarycentricCoordinates4 for tetrahedrons

Andreas-Forster:
This PR adds the class BarycentricCoordinates4. The class represents the barycentric coordinates (bc) for a tetrahedron. With this class, the bc calculations can be made entirely in scala which is faster than accessing VTK.

Tests also cover that the calculation returns the same result as VTK, and that the calculation is faster than using VTK.

### DIFF
--- a/src/main/scala/scalismo/mesh/BarycentricCoordinates.scala
+++ b/src/main/scala/scalismo/mesh/BarycentricCoordinates.scala
@@ -111,3 +111,102 @@ object BarycentricCoordinates {
       new BarycentricCoordinates(1.0 - s, 1.0 - t, s + t - 1.0)
   }
 }
+
+/** barycentric coordinates for localization within a tetrahedron */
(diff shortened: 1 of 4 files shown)


Andreas-Forster:
Corrected formating with force-push.

Andreas-Forster:
Would you deprecate and then later remove `BarycentricCoordinates`? I am not so in favor of having them both used in the future.

I am also undecided if it is necessary to rename it. I think when you learn our library, people start with triangles first and will use `BarycentricCoordinates`, kind of natural and easy. When you move later to tetrahedrons then it is `BarycentricCoordinates4`.

marcelluethi:
We don't have to take a decision now, but can keep it for the moment the way it is. Whenever we think the inconsistency causes problems we can still rename/deprecate. 

My hope is, of course, that tetrahedral meshes will be the new killer feature for which people come to scalismo. Then they would first be exposed to ```BarycentricCoordinates4``` 😄

PR #322, opened 2020-02-04
adding closest point calculations for tetrahedral mesh with test cove…

Andreas-Forster:
This PR adds to the tetrahedral mesh operations the query for the closest point. Additional tests cover the most important parts of the added functionality.

Behavior:
- For all points outside of the mesh, it returns the distance to the closest point.
- For points within one of the tetrahedrons, it returns the distance 0.

### DIFF
--- a/src/main/scala/scalismo/mesh/MeshOperations.scala
+++ b/src/main/scala/scalismo/mesh/MeshOperations.scala
@@ -54,7 +54,7 @@ class TriangleMesh3DOperations(private val mesh: TriangleMesh3D) {
   def shortestDistanceToSurfaceSquared(point: Point[_3D]): Double =
     closestPointOnSurface.getSquaredShortestDistance(point: Point[_3D])
   def closestPoint(point: Point[_3D]): ClosestPoint = closestPointOnSurface.getClosestPoint(point)
-  def closestPointOnSurface(point: Point[_3D]): ClosestPointOnSurface =
+  def closestPointOnSurface(point: Point[_3D]): ClosestPointWithSquaredDistance =
     closestPointOnSurface.getClosestPointOnSurface(point)
 
   /**
@@ -226,6 +226,14 @@ class TetrahedralMesh3DOperations(private val mesh: TetrahedralMesh[_3D]) {
    */
   private lazy val tetrahedrons = BoundingSpheres.tetrahedronListFromTetrahedralMesh3D(mesh)
   private lazy val boundingSpheres = BoundingSpheres.createForTetrahedrons(tetrahedrons)
+
+  private lazy val closestPointIndex: VolumeSpatialIndex[_3D] =
+    new TetrahedralMesh3DSpatialIndex(boundingSpheres, mesh, tetrahedrons)
+  def shortestDistanceToVolumeSquared(point: Point[_3D]): Double =
+    closestPointIndex.getSquaredShortestDistance(point: Point[_3D])
+  def closestPoint(point: Point[_3D]): ClosestPoint = closestPointIndex.getClosestPoint(point)
+  def closestPointToVolume(point: Point[_3D]): ClosestPointWithSquaredDistance =
+    closestPointIndex.getClosestPointToVolume(point)
 
   private lazy val intersect: TetrahedralizedVolumeIntersectionIndex[_3D] =
     new LineTetrahedralMesh3DIntersectionIndex(boundingSpheres, mesh, tetrahedrons)
--- a/src/main/scala/scalismo/mesh/MeshOperations.scala
+++ b/src/main/scala/scalismo/mesh/MeshOperations.scala
@@ -54,7 +54,7 @@ class TriangleMesh3DOperations(private val mesh: TriangleMesh3D) {
   def shortestDistanceToSurfaceSquared(point: Point[_3D]): Double =
     closestPointOnSurface.getSquaredShortestDistance(point: Point[_3D])
   def closestPoint(point: Point[_3D]): ClosestPoint = closestPointOnSurface.getClosestPoint(point)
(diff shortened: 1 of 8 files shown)


src/main/scala/scalismo/mesh/boundingSpheres/SurfaceSpatialIndex.scala
```diff
@@ -39,24 +43,33 @@ trait SurfaceSpatialIndex[D] {
...
-private object ClosestPointType extends Enumeration {
-  type ClosestPointType = Value
-  val POINT, ON_LINE, IN_TRIANGLE = Value
+object SurfaceSpatialIndex {
+
+  /**
+   * Type of the closest point. At the moment the names are only suited for a triangular mesh.
+   */
+  private[boundingSpheres] object SurfaceClosestPointType extends Enumeration {
+    type SurfaceClosestPointType = Value
```
marcelluethi: Maybe ClosestPointType would be sufficient? It is clear from the surrounding context that it refers to the ClosestPointType of the surface.

src/main/scala/scalismo/mesh/boundingSpheres/ClosestPoint.scala
```diff
@@ -36,9 +36,9 @@ case class ClosestPoint(point: Point[_3D], distanceSquared: Double) {
  * @param point The closest point location on the surface.
  * @param distanceSquared The squared distance to the closest point location.
  */
-sealed abstract class ClosestPointOnSurface(val point: Point[_3D], val distanceSquared: Double) {
+sealed abstract class ClosestPointWithSquaredDistance(val point: Point[_3D], val distanceSquared: Double) {
```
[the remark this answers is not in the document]
Andreas-Forster: The `ClosestPoint` class is for the most performant queries where no meta-information (isVertex, isOnLine, isInTriangle etc.) is needed. As it is a case class, it can not serve as the base class for the other closest point classes with more information. `ClosestPointWithMetaInformation` or `ClosestPointBase` would be a more appropriate name. But I do not like those names. Somehow the name should make it clear that one can get more information when one matches its concrete type.

src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -227,6 +227,14 @@ class TetrahedralMesh3DOperations(private val mesh: TetrahedralMesh[_3D]) {
   private lazy val tetrahedrons = BoundingSpheres.tetrahedronListFromTetrahedralMesh3D(mesh)
   private lazy val boundingSpheres = BoundingSpheres.createForTetrahedrons(tetrahedrons)
 
+  private lazy val closestPointIndex: VolumeSpatialIndex[_3D] =
+    new TetrahedralMesh3DSpatialIndex(boundingSpheres, mesh, tetrahedrons)
+  def shortestDistanceToVolumeSquared(point: Point[_3D]): Double =
```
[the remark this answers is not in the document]
Andreas-Forster: This name is kept as similar as possible to the `shortestPointToSurfaceSquared` which exists in the `TriangleMesh3DOperations`.

src/main/scala/scalismo/mesh/boundingSpheres/BSDistance.scala
```diff
@@ -92,7 +94,7 @@ case class Tetrahedron(a: EuclideanVector[_3D],
 
     var larg = (0, 0.0)
 
-    val largestFace = for (i <- 0 to sq.size - 1) {
+    for (i <- 0 to sq.size - 1) {
```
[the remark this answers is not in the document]
Andreas-Forster: Here I just removed the variable, as it is never used. This reduces a warning depending on the specified compiler options. I think we can deprecate anyway this function here. The class is meant to be used only in the spatial index classes and there it is not used. So I do not like to spend time rewriting it.

src/main/scala/scalismo/mesh/boundingSpheres/SurfaceSpatialIndex.scala
Andreas-Forster: There are two versions of those. `SurfaceClosestPointType` and `VolumeClosestPointType` exists both. But I will remove the surrounding object which is not needed.

Andreas-Forster: Object around removed but kept the name.

src/main/scala/scalismo/mesh/boundingSpheres/ClosestPoint.scala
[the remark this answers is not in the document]
Andreas-Forster: I now changed the name to `ClosestPointSpecialized`. While I am still not totally happy with that name, I think it comes closest to what it should express.

Further, I added the method `distance` that returns the square root of the `squaredDistance`.

src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -54,7 +54,7 @@ class TriangleMesh3DOperations(private val mesh: TriangleMesh3D) {
   def shortestDistanceToSurfaceSquared(point: Point[_3D]): Double =
     closestPointOnSurface.getSquaredShortestDistance(point: Point[_3D])
   def closestPoint(point: Point[_3D]): ClosestPoint = closestPointOnSurface.getClosestPoint(point)
-  def closestPointOnSurface(point: Point[_3D]): ClosestPointOnSurface =
+  def closestPointOnSurface(point: Point[_3D]): ClosestPointSpecialized =
```
marcelluethi: How about ```ClosestPointWithType``` instead of specialized? Later on you use the name ClosestPointType for this distinction.

Andreas-Forster: Thank you for this suggestion. I like this name.

PR #324, opened 2020-02-06
Add surface predicate for tetrahedral meshes

Andreas-Forster:
This PR should be merged after merging #322 .

The PR adds the boundary predicate to `TetrahedralMesh3D` in the same way as it exists for `TriangleMesh3D`. It also lets you extract the surface of the tetrahedral mesh as a `TriangleMesh3D`.

Tests for the tetrahedron and point predicate are added.


Andreas-Forster:
Replaced by #325

PR #325, opened 2020-02-06
added boundary predicates for tetrahedral meshes

Andreas-Forster:
This PR replaces #324 

The PR adds the boundary predicate to TetrahedralMesh3D in the same way as it exists for TriangleMesh3D. It also lets you extract the surface of the tetrahedral mesh as a TriangleMesh3D.

Tests for the tetrahedron and point predicate are added.

### DIFF
--- a/src/main/scala/scalismo/mesh/MeshBoundaryPredicates.scala
+++ b/src/main/scala/scalismo/mesh/MeshBoundaryPredicates.scala
@@ -17,10 +17,7 @@ package scalismo.mesh
 
 import breeze.linalg
 import breeze.linalg.CSCMatrix
-import org.junit.runner.manipulation.Sortable
 import scalismo.common.PointId
-
-import scala.collection.mutable
 
 /**
  * MeshBoundary is a property to test if points or an edge between two points is on the boundary of the mesh.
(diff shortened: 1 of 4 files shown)


src/test/scala/scalismo/mesh/boundingSpheres/MeshSurfaceDistanceTests.scala
```diff
@@ -595,7 +595,6 @@ class MeshSurfaceDistanceTests extends ScalismoTestSuite {
       def isInside(pt: Point[_3D]) = {
         tmesh.tetrahedrons.exists { cell =>
           val bc = tmesh.getBarycentricCoordinates(pt, cell)
-          println()
```
Andreas-Forster: Removed `println()` statement which made it accidentally to the tests.

src/main/scala/scalismo/mesh/MeshBoundaryPredicates.scala
```diff
@@ -15,9 +15,13 @@
  */
 package scalismo.mesh
 
+import breeze.linalg
 import breeze.linalg.CSCMatrix
+import org.junit.runner.manipulation.Sortable
```
marcelluethi: I suppose this import is not needed.

src/main/scala/scalismo/mesh/MeshBoundaryPredicates.scala
```diff
@@ -77,11 +96,47 @@ private class BoundaryOfATriangleMeshPredicates(private var vertexIsOnBorder: In
   }
 }
 
+/**
+ * Implementation of a TetrahedralMeshBoundary
+ */
+private class BoundaryOfATetrahedralMeshPredicates(private var vertexIsOnBorder: IndexedSeq[Boolean],
```
marcelluethi: do these really need to be vars or would vals also do it?

src/main/scala/scalismo/mesh/MeshBoundaryPredicates.scala
Andreas-Forster: Remove in newest version.

src/main/scala/scalismo/mesh/MeshBoundaryPredicates.scala
Andreas-Forster: Changed to vals.

PR #327, opened 2020-02-12
moved StatisticalVolumeMeshModels (with io) to experimental

marcelluethi:
In the current version of the software, there is no unified concept of meshmodels and volumemeshmodels, even though they are closely related. I have started working on such a unification, but it will need quite some time and testing before it can be released. 

In order not to block things, I propose to go ahead with this version, but to mark the parts which are likely to be affected by the redesign as experimental. To achieve this, I have moved StatisticalVolumeMeshModes and IntensityModels along with the corresponding IO classes into a package ```experimental```.

@kahrpatrick @rassaire Could you quickly comment whether this is okay for you like this?


Andreas-Forster:
I am unsure if it is worth doing this step to place stuff in experimental. I do not really see why we should make the SVMM experimental while the SMM might also change when harmonizing stuff. Otherwise, it would suggest that we should keep the SMM interface while it might be better to adapt it when we see the common parts/concepts with SVMM.

So depending on your time estimate I would prefer to either have another milestone release or later on a breaking release that changes the interfaces. But I could also live with the experimental path/way to introduce SVMM.

rassaire:
If I understand correctly the experimental package will be in milston2? while milstone1 will remains as it is?

marcelluethi:
Thanks for the feedback. 

For me it is all about communication and managing expectation. If we release it unchanged, it appears as if this would be something stable. If we have it in experimental, we are explicit about the fact that the users will need to adapt their code if they use it. We might even get some feedback during that time. 

The SMM have been around for quite some time and were stable for 4 years. There the correct path is to deprecate them and let users gradually transition to the new concept.

There are plenty of things in the new release that are already useful. I would not want to delay it. Also the FutureLearn course and our own shape modelling course starts next week. It would be great if we had at least a release candidate ready.

PR #331, opened 2020-02-13
update BarycentricInterpolator to use MeshOperations

kahrpatrick:
This PR removes a temporary solution to find the corresponding mesh cell for a point. Instead it makes use of the recently added MeshOperations for tetrahedral meshes. The consequence is that, in contrast to the previous version, points outside of the mesh will be interpolated using a nearest neighbor scheme. The modifications in this PR should also make the interpolation more efficient.

### DIFF
--- a/src/main/scala/scalismo/common/interpolation/BarycentricInterpolator.scala
+++ b/src/main/scala/scalismo/common/interpolation/BarycentricInterpolator.scala
@@ -1,10 +1,10 @@
 package scalismo.common.interpolation
 
 import scalismo.common._
-import scalismo.geometry.{_3D, NDSpace, Point, Point3D}
-import scalismo.mesh.{TetrahedralCell, TetrahedralMesh, TetrahedronId}
+import scalismo.geometry.{_3D, NDSpace, Point}
+import scalismo.mesh.boundingSpheres._
+import scalismo.mesh.{MeshOperations, TetrahedralMesh}
 import scalismo.numerics.ValueInterpolator
-import scalismo.utils.Memoize
 
 trait BarycentricInterpolator[D, A] extends FieldInterpolator[D, UnstructuredPointsDomain[D], A] {
   implicit protected val valueInterpolator: ValueInterpolator[A]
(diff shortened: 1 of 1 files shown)


src/main/scala/scalismo/common/interpolation/BarycentricInterpolator.scala
```diff
@@ -33,48 +33,33 @@ object BarycentricInterpolator {
...
-    cell
-  }
+  override protected val valueInterpolator: ValueInterpolator[A] = ValueInterpolator[A]
 
   override def interpolate(df: DiscreteField[_3D, UnstructuredPointsDomain[_3D], A]): Field[_3D, A] = {
 
     def interpolateBarycentric(p: Point[_3D]): A = {
-      getTetrahedralMeshCell(p) match {
-        case Some(cell) =>
+
```
Andreas-Forster: I would find something like this more clear (mainly because of the last to cases:
```
    cp match {
      case ClosestPointIsVertex(_, _, pid)    => df(pid)
      case ClosestPointOnLine(_, _, pids, bc) => ValueInterpolator[A].blend(df(pids._1), df(pids._2), bc)
      case ClosestPointInTriangleOfTetrahedron(_, _, tetId, triId, bc) => {
        val triangle = mesh.tetrahedralization.tetrahedron(tetId).triangles(triId.id)
        bc.interpolateProperty(df(triangle.ptId1), df(triangle.ptId2), df(triangle.ptId3))
      }
      case ClosestPointInTetrahedron(_, _, tid, bc) => {
        val tetrahedron = mesh.tetrahedralization.tetrahedron(tid)
        bc.interpolateProperty(df(tetrahedron.ptId1), df(tetrahedron.ptId2), df(tetrahedron.ptId3), df(tetrahedron.ptId4))
      }
    }
```

src/main/scala/scalismo/common/interpolation/BarycentricInterpolator.scala
```diff
@@ -33,48 +33,33 @@ object BarycentricInterpolator {
...
-  private def getTetrahedralMeshCell(p: Point[_3D]): Option[TetrahedralCell] = {
-
-    val numberOfTetrahedrons = mesh.tetrahedralization.tetrahedrons.length
-
-    def isInsideCell(tc: TetrahedralCell): Boolean = mesh.isInsideTetrahedralCell(p, tc)
-    val isInsideCellMemoized = Memoize(isInsideCell, numberOfTetrahedrons)
-
-    var cell: Option[TetrahedralCell] = None
-    var neighbourhood = Set[TetrahedronId]()
+  val meshOps = MeshOperations(mesh)
```
Andreas-Forster: Is there a reason why not to use `mesh.operations`?

src/main/scala/scalismo/common/interpolation/BarycentricInterpolator.scala
kahrpatrick: I agree, this looks cleaner.

src/main/scala/scalismo/common/interpolation/BarycentricInterpolator.scala
kahrpatrick: Not really, will change this.

PR #338, opened 2020-03-10
don't throw exception when ply file does not exist. Fixes #337

### LINKED ISSUE #337: ply reader throws exception when file is not found
The PLY reader throws an exception if a file is not found. The correct behavior would be to return a Failure instead. 

The problem can be reproduced with the following statement:
```scala
val reader = MeshIO.readMesh(new java.io.File("donotexist.ply"))
```
which throws an exception, even though ```.get``` is not called.

marcelluethi:
This PR corrects the bug that the MeshIO method would throw an exception when a given .ply file does not exist instead of simply returning a failure.

### DIFF
--- a/src/main/scala/scalismo/io/MeshIO.scala
+++ b/src/main/scala/scalismo/io/MeshIO.scala
@@ -596,7 +596,7 @@ object MeshIO {
       if (headerLines.exists(_.contains("TextureFile")) && headerLines.exists(_.contains("format ascii"))) {
         Failure(
           new IOException(
-            "PLY file seems to be a textured mesh in ASCII format which creates issues with the VTK ply reader. Please convert it to a binary ply or to a vertex color or shape only ply."
+            "PLY file $filename seems to be a textured mesh in ASCII format which creates issues with the VTK ply reader. Please convert it to a binary ply or to a vertex color or shape only ply."
           )
         )
       } else {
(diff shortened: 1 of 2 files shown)


src/main/scala/scalismo/io/MeshIO.scala
```diff
@@ -583,58 +583,64 @@ object MeshIO {
...
-      plyReader.Update()
+      val headerLines = lineIterator.dropWhile(_ != "ply").takeWhile(_ != "end_header").toIndexedSeq
 
-      val errCode = plyReader.GetErrorCode()
-      if (errCode != 0) {
-        return Failure(new IOException(s"Could not read ply mesh (received VTK error code $errCode"))
+      if (headerLines.exists(_.contains("TextureFile")) && headerLines.exists(_.contains("format ascii"))) {
+        Failure(
+          new IOException(
+            "PLY file seems to be a textured mesh in ASCII format which creates issues with the VTK ply reader. Please convert it to a binary ply or to a vertex color or shape only ply."
```
Andreas-Forster: Add filename to the error message as for the missing file case?

src/main/scala/scalismo/io/MeshIO.scala
```diff
@@ -583,58 +583,64 @@ object MeshIO {
...
-            }
-            Right(VertexColorMesh3D(meshGeometry, new SurfacePointProperty[RGBA](meshGeometry.triangulation, colors)))
+  private def readPLYUsingVTK(file: File): Try[Either[TriangleMesh[_3D], VertexColorMesh3D]] = {
+    val plyReader = new vtkPLYReader()
+    plyReader.SetFileName(file.getAbsolutePath)
+    plyReader.Update()
+
+    val errCode = plyReader.GetErrorCode()
+    if (errCode != 0) {
+      return Failure(new IOException(s"Could not read ply mesh (received VTK error code $errCode"))
```
Andreas-Forster: Add filename to the error message as for the missing file case?

src/main/scala/scalismo/io/MeshIO.scala
marcelluethi: Thanks for the review. This is a good point. I have addressed it in the last commit.

PR #339, opened 2020-03-23
Added missing copyright information

BernhardEgger:
I just found that this file is missing  license information
Perhaps there are more files like that


marcelluethi:
Thanks for adding this.

PR #341, opened 2020-04-02
PointDistributionModel IO method

madsendennis:
PointDistributionModel IO method implementation for the new domain concept.


src/test/scala/scalismo/io/StatisticalModelIOTest.scala
```diff
@@ -59,33 +60,33 @@ class StatisticalModelIOTest extends ScalismoTestSuite {
...
-        model <- StatismoIO.readStatismoMeshModel(statismoFile)
-        _ <- StatismoIO.writeStatismoMeshModel(model, dummyFile, statismoVersion = v081)
-        readModel <- StatismoIO.readStatismoMeshModel(dummyFile)
-      } yield {
-        assertModelAlmostEqual(model, readModel)
-      }
-      t.get
-
-    }
+//    it("can be written in version 0.81 and read again") {
```
marcelluethi, line 73: We should just remove this test instead of commenting it out.

madsendennis:
Added the corrections and added support for reading a model with older version number (0.8).

marcelluethi:
Great. We are getting there.

There is still the separate PointDistributionIO object. My suggestion would be to get rid of it and move the method into the StatisticalModelIO. There the user can find all the statistical model methods, including those for reading intensitymodels (which are not pdms). Or would you prefer to separate each type of model?

madsendennis:
I've now added a few tests to the StatismoDomainIO and some general PDM io tests.
Currently all PDMs are written with the _datasetType_ set to _POLYGON_MESH_ , would it make sense to  write the mesh connectivity into the model itself?

marcelluethi:
Yes, the type should be written into the model. This is specified in the statismo format. The corresponding constant that should be read/written can easily be specified in the typeclass.

src/main/scala/scalismo/io/ActiveShapeModelIO.scala
```diff
@@ -56,8 +56,10 @@ object ActiveShapeModelIO {
   }
 
   def writeActiveShapeModel(asm: ActiveShapeModel, file: File): Try[Unit] = {
+
+    val pointModel = PointDistributionModel[_3D, TriangleMesh](asm.statisticalModel.gp)
     for {
-      _ <- StatismoIO.writeStatismoMeshModel(asm.statisticalModel, file)
+      _ <- StatismoIO.writeStatismoPointModel(pointModel, file)
```
marcelluethi, line 62: I would call it writeStatismoPDM instead of PointModel

src/main/scala/scalismo/io/ActiveShapeModelIO.scala
```diff
@@ -100,7 +102,7 @@ object ActiveShapeModelIO {
 
   def readActiveShapeModel(fn: File): Try[ActiveShapeModel] = {
     for {
-      shapeModel <- StatismoIO.readStatismoMeshModel(fn)
+      pdm <- StatismoIO.readStatismoPointModel[_3D, TriangleMesh](fn)
```
marcelluethi, line 105: Same here, readStatismoPDM

madsendennis: fixed

src/main/scala/scalismo/io/ActiveShapeModelIO.scala
madsendennis: fixed

PR #343, opened 2020-04-24
Added support for Scala 2.13

marcelluethi:
Added support for Scala 2.13. Remained backward compatible with 2.12, but dropped support for 2.11.


marcelluethi:
@Andreas-Forster @madsendennis Any objections to merging this PR?

build.sbt
```diff
@@ -7,8 +7,8 @@ lazy val root = (project in file("."))
   .settings(
     name := "scalismo",
     organization := "ch.unibas.cs.gravis",
-    scalaVersion := "2.12.6",
-    crossScalaVersions := Seq("2.11.12", "2.12.6"),
+    scalaVersion := "2.13.2",
+    crossScalaVersions := Seq("2.12.11", "2.13.2"),
```
Andreas-Forster, line 11: Is the support for 2.11 no longer needed or is only two minor versions our rule of thumb?

marcelluethi:
I think we could still support Scala 2.11 but I have not checked. As this release will be a rather big transition anyway, I would be in favor of cutting as many old branches as possible and that we develop without being constraint to support very old versions. 

Having said that, I am in favor of adding 2.11 again if we see that it all works without requiring code changes.

madsendennis:
Seems good to me.

Issue #344, opened 2020-04-28
imageBoundingBox inconsistencies


kahrpatrick:
Let `domain: DiscreteImageDomain3D` be the domain of an image. Then `domain.imageBoundingBox` represents the rectangular region over which the image is defined. It is 1*spacing larger than `domain.boundingBox`.

Issues:

1) Interpolating the image using the `BSplineImageInterpolator3D`, returns a field that is only defined inside `domain.boundingBox` instead of `domain.imageBoundingBox`.

2) It is possible to find `point: Point3D` for which `domain.isDefinedAt(point)` returns `false` while `domain.imageBoundingBox.isDefinedAt(point)` returns `true`. This should not be the case.

To reproduce, take `domain.boundingBox.oppositeCorner` and gradually add 0-1 times `domain.spacing`. This should give points for which the image is still defined.

Code to reproduce issues:
```
  val image = ImageIO.read3DScalarImage[Float](new File("/src/test/resources/images/vtk/3d_float.vtk")).get
  val domain = image.domain
  val bSplineInterpolator = BSplineImageInterpolator3D[Float](3)
  val imageCont = image.interpolate(bSplineInterpolator)

  for(nextDouble <- 0.0 to 1.0 by 0.1){
    val samplePoint = domain.boundingBox.oppositeCorner+domain.spacing*nextDouble

    println(s"domain is defined at $samplePoint : ${domain.isDefinedAt(samplePoint)}")
    println(s"boundingBox is defined at $samplePoint : ${domain.boundingBox.isDefinedAt(samplePoint)}")
    println(s"imageBoundingBox is defined at $samplePoint : ${domain.imageBoundingBox.isDefinedAt(samplePoint)}")
    println(s"continuous image is defined at $samplePoint : ${imageCont.isDefinedAt(samplePoint)}")
  }
```

Output:
```
domain is defined at Point3D(1.0,1.0,1.0) : true
boundingBox is defined at Point3D(1.0,1.0,1.0) : true
imageBoundingBox is defined at Point3D(1.0,1.0,1.0) : true
continuous image is defined at Point3D(1.0,1.0,1.0) : true

domain is defined at Point3D(1.1,1.1,1.1) : false
boundingBox is defined at Point3D(1.1,1.1,1.1) : false
imageBoundingBox is defined at Point3D(1.1,1.1,1.1) : true
continuous image is defined at Point3D(1.1,1.1,1.1) : false

[...]

PR #345, opened 2020-04-29
use imageBoundingBox instead of boundingBox

kahrpatrick:
Fixes the first part of #344 by using the imageBoundingBox instead of boundingBox for the interpolated image.


marcelluethi:
Thanks for the fix. 
I changed the target branch to release-0.18, as this is a bug in the production version. Once it is merged we can merge release-0.18 into develop to also have it available there.

PR #348, opened 2020-05-14
Image cleanup

marcelluethi:
During the refactoring to the new domain concept, the image classes were adapted only to the point that everything works, but not such that the concept is consistent. This PR does makes this now more consistent. In particular, the classes ```DiscreteImageDomain``` and ```StructuredPoints``` were adapted. Also the concept of  a ```DiscreteScalarImage``` was removed, and replaced with the type ```DiscreteImage```, which in turn is just a type alias for ```DiscreteField[D, DiscreteImageDomain, A]```.

To illustrate the changes, a tutorial introduction was added (see the docs/images.md). A dependency on mdoc was added to sbt, in order to be able to compile the tutorial document.


marcelluethi:
@kahrpatrick @Andreas-Forster Have you had a chance to look into this PR and see if you are okay with the new interface? If yes I would like to do a final cleanup round and merge it into develop, such that we can move on. There will be a new set of changes regarding transformations coming soon, which depends on this PR.

kahrpatrick:
> @kahrpatrick @Andreas-Forster Have you had a chance to look into this PR and see if you are okay with the new interface? If yes I would like to do a final cleanup round and merge it into develop, such that we can move on. There will be a new set of changes regarding transformations coming soon, which depends on this PR.

Although I did not thoroughly test the code, I could not identify any issues with the proposed refactoring. Looking forward to seeing the new changes.

src/main/scala/scalismo/image/DiscreteImageDomain.scala
```diff
@@ -0,0 +1,196 @@
...
+  override def pointSet: StructuredPoints[D] = structuredPoints
+
+  def origin = pointSet.origin
+  def spacing = pointSet.spacing
+  def size = pointSet.size
+
+  def boundingBox: BoxDomain[D] = {
+
+    // The image bounding box is 1*spacing larger than the bounding box of the point of the domain, as
+    // every point of the domain represents one voxel.
```
Andreas-Forster, line 48: might be important to mention here that this implementation implies that for each voxel, the point it defines is at the corner to the negative side of each axis. Or is this convention common sense?

src/main/scala/scalismo/image/DiscreteScalarImage.scala
```diff
@@ -31,7 +31,7 @@
 // * @tparam D  The dimensionality of the image
 // * @tparam A The type of the pixel (needs to implement Scalar).
 // */
-//class DiscreteScalarImage[D: NDSpace, A: Scalar: ClassTag](override val domain: DiscreteImageDomain[D],
+//class DiscreteImage[D: NDSpace, A: Scalar: ClassTag](override val domain: DiscreteImageDomain[D],
```
Andreas-Forster, line 34: remove commented code? (also a lot following in the same file)

src/main/scala/scalismo/image/Image.scala
```diff
@@ -75,7 +74,7 @@ import scala.reflect.ClassTag
 //   * support size of the filter and the Frequency of the image, increasing this value can help avoid artifacts (at the cost of heavier computation)
 //   */
 //  def convolve(filter: Filter[D],
-//               numberOfPointsPerDim: Int)(implicit c: CreateDiscreteImageDomain[D]): ScalarImage[D, A] = {
+//               numberOfPointsPerDim: Int)(implicit c: CreateStructuredPoints[D]): ScalarImage[D, A] = {
```
Andreas-Forster, line 77: remove commented code? (also more commented code in this file)

src/main/scala/scalismo/image/DiscreteImageDomain.scala
marcelluethi: Thanks. Added the following clarification:
```
// (The voxel that is defined by each grid point extends by the length spacing(i) into the
// i-th space direction).
```

src/main/scala/scalismo/image/DiscreteScalarImage.scala
marcelluethi: I will delete all the code that was commented out.  In most cases I could just remove the source file itself.

PR #354, opened 2020-08-14
parallelized discretization of fields

marcelluethi:
Added parallelisation of discretization of fields. This functionality was already there for images in the previous version of scalismo, but was removed in the redesign. Now it is reintroduced for more general fields.

### DIFF
--- a/src/main/scala/scalismo/common/UnstructuredPoints.scala
+++ b/src/main/scala/scalismo/common/UnstructuredPoints.scala
@@ -66,8 +66,8 @@ sealed abstract class UnstructuredPoints[D: NDSpace: Create] private[scalismo] (
   }
 
   override def pointsInChunks(nChunks: Int): Seq[Iterator[Point[D]]] = {
-    val chunkSize = pointSequence.size / nChunks
-    val chunks = for (i <- 0 until (nChunks + 1)) yield {
+    val chunkSize = Math.ceil(pointSequence.size / nChunks.toDouble).toInt
+    val chunks = for (i <- 0 until nChunks) yield {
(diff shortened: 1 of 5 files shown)


src/main/scala/scalismo/common/UnstructuredPoints.scala
```diff
@@ -65,6 +65,14 @@ sealed abstract class UnstructuredPoints[D: NDSpace: Create] private[scalismo] (
 
   }
 
+  override def pointsInChunks(nChunks: Int): Seq[Iterator[Point[D]]] = {
+    val chunkSize = pointSequence.size / nChunks
+    val chunks = for (i <- 0 until (nChunks + 1)) yield {
```
Andreas-Forster, line 70: Could it be that this returns you n+1 chunks?

src/main/scala/scalismo/common/UnstructuredPoints.scala
```diff
@@ -65,6 +65,14 @@ sealed abstract class UnstructuredPoints[D: NDSpace: Create] private[scalismo] (
 
   }
 
+  override def pointsInChunks(nChunks: Int): Seq[Iterator[Point[D]]] = {
+    val chunkSize = pointSequence.size / nChunks
```
Andreas-Forster, line 69: Shouldn't this be floating point arithmetics and then ceiled? Otherwise, you will have to gather the last few elements in an additional chunk.

marcelluethi: you are totally right. This was a beginner's mistake. Should be fixed now

PR #360, opened 2020-12-28
Introduced SampleLense and added GaussianRandomWalkProposal

marcelluethi:
This commit introduced the trait ```SampleLense``` , which is a general way to access parts of a generic sample for the mcmc framework. This mechanism makes it possible to keep the samples completely generic, but still be able to provide useful proposal generators. As an example, the ```GaussianRandomWalkProposal``` is introduced. While this might seems like an overkill for this simple proposal generator, a lot of errors can be avoided when we  introduce more complicated proposals. Indeed, the motivation for this is the Mala proposal, which will be introduced in a separate PR:

The following code shows a complete example of how the lense can be used. Proposals are generated blockwise, where each parameter (mu, sigma) is updated separatly using the ```GaussianRandomWalkProposal```. 

```scala
import breeze.linalg.DenseVector
import scalismo.sampling.algorithms.MetropolisHastings
import scalismo.sampling.evaluators.ProductEvaluator
import scalismo.sampling.proposals.{GaussianRandomWalkProposal, MalaProposal, MixtureProposal}
import scalismo.sampling.{DistributionEvaluator, GradientEvaluator, SampleLens}

// Parameter class - This is generic and every application chooses its
// own structures of the parameters
case class Parameters(mu: Double, sigma: Double)
case class Sample(parameters: Parameters, generatedBy: String)

// The lense for accessing the mean from the parameter vector
object MeanParamLens extends SampleLens[Sample] {

  override def numberOfParameters: Int = 1

  override def getAsVector(sample: Sample): DenseVector[Double] = {
    val v = DenseVector.zeros[Double](1)
    v(0) = sample.parameters.mu
    v
  }
  override def setFromVector(sample: Sample, vector: DenseVector[Double], generatedBy: Option[String]): Sample = {
    sample.copy(parameters = sample.parameters.copy(mu = vector(0)))
  }
}

// The lense for accessing the staddev from the parameter vector
object StddevParamLens extends SampleLens[Sample] {

  override def numberOfParameters: Int = 1

  override def getAsVector(sample: Sample): DenseVector[Double] = {
    val v = DenseVector.zeros[Double](1)
    v(0) = sample.parameters.sigma
    v
  }

  override def setFromVector(sample: Sample, vector: DenseVector[Double], generatedBy: Option[String]): Sample = {
    sample.copy(parameters = sample.parameters.copy(sigma = vector(0)))
  }
}

case class Evaluator(data: Seq[Double]) extends DistributionEvaluator[Sample] {

  override def logValue(theta: Sample): Double = {
    val likelihood = breeze.stats.distributions.Gaussian(
      theta.parameters.mu,
      theta.parameters.sigma
    )
    val likelihoods = for (x <- data) yield {
      likelihood.logPdf(x)
    }
    likelihoods.sum
  }
}

object MinimalExample {

  def generateSyntheticData(numDataPoints: Int): Seq[Double] = {
    // The true distribution to sample from
    val mu = -1
    val sigma = 3

    val trueDistribution = breeze.stats.distributions.Gaussian(mu, sigma)
    for (_ <- 0 until numDataPoints) yield {
      trueDistribution.draw()
    }
  }

  def main(args: Array[String]): Unit = {

    scalismo.initialize()
    implicit val rng = scalismo.utils.Random(42)

    val data = generateSyntheticData(100)
    val evaluator = Evaluator(data)

    // Using the new Mala proposal as a gradient
    val generatorForMean = GaussianRandomWalkProposal(1, MeanParamLens)
    val generatorForStddev = GaussianRandomWalkProposal(0.1, StddevParamLens)
    val generator = MixtureProposal.fromProposalsWithTransition((0.5, generatorForMean), (0.5, generatorForStddev))

    val chain = MetropolisHastings(generator, evaluator)

    val initialSample = Sample(Parameters(0, 1), generatedBy = "initial")
    val mhIterator = chain.iterator(initialSample)
    val samples = mhIterator.drop(1000).take(100000).toIndexedSeq

    val estimatedMean = samples.map(sample => sample.parameters.mu).sum / samples.size
    val estimatedSigma = samples.map(sample => sample.parameters.sigma).sum / samples.size

    println((estimatedMean, estimatedSigma))
  }
}
```

### DIFF
--- a/src/main/scala/scalismo/sampling/SampleLens.scala
+++ b/src/main/scala/scalismo/sampling/SampleLens.scala
@@ -1,0 +1,33 @@
+package scalismo.sampling
+
+import breeze.linalg.DenseVector
+
+/**
+ * This trait is used in the context of an MCMC chain, where the we have a class Sample or
+ * generic type A, which contains an arbitrary number of parameters, with arbitrary structure.
+ * The lens focuses attention to one part of the parameter vector, which can then be read and updated
+ * using the provided getter and setter methods.
+ * To update different parts of a sample, different lenses must be implemented
+ */
+trait SampleLens[A] {
+
+  /**
+   * The number of parameters this lense focuses on
+   */
+  def numberOfParameters: Int
+
+  /**
+   * returns the part of the parameters on which this lense focuses as a dense vector
+   * The length of the vectors needs to correspond to numberOfParameters
+   */
+  def getAsVector(sample: A): DenseVector[Double]
+
+  /**
+   * returns a new sample, which the part, on which this lense focuses replaced by the values
+   * given in the vector. The option generatedBy argument is used to tag who has generated
+   * the resulting sample.
+   *
+   * The length of the given vectors needs to correspond to numberOfParameters
+   */
+  def setFromVector(sample: A, vector: DenseVector[Double], generatedBy: Option[String] = None): A
+}
--- a/src/main/scala/scalismo/sampling/SampleLens.scala
+++ b/src/main/scala/scalismo/sampling/SampleLens.scala
@@ -3,31 +3,25 @@ package scalismo.sampling
 import breeze.linalg.DenseVector
 
 /**
- * This trait is used in the context of an MCMC chain, where the we have a class Sample or
- * generic type A, which contains an arbitrary number of parameters, with arbitrary structure.
- * The lens focuses attention to one part of the parameter vector, which can then be read and updated
+ * This class is used in the context of an MCMC chain, where the we have a class Sample or
+ * generic type A, with arbitrary structure.
+ * The lens focuses attention to a part of the sample with type B, which can then be read and updated
  * using the provided getter and setter methods.
  * To update different parts of a sample, different lenses must be implemented
  */
-trait SampleLens[A] {
+abstract class SampleLens[A, B] {
 
   /**
-   * The number of parameters this lense focuses on
+   * returns the part of the parameters on which this lens focuses
    */
-  def numberOfParameters: Int
+  def get(sample: A): B
 
   /**
-   * returns the part of the parameters on which this lense focuses as a dense vector
-   * The length of the vectors needs to correspond to numberOfParameters
-   */
-  def getAsVector(sample: A): DenseVector[Double]
-
-  /**
-   * returns a new sample, which the part, on which this lense focuses replaced by the values
+   * returns a new sample, which the part, on which this lens focuses replaced by the values
    * given in the vector. The option generatedBy argument is used to tag who has generated
    * the resulting sample.
    *
    * The length of the given vectors needs to correspond to numberOfParameters
    */
-  def setFromVector(sample: A, vector: DenseVector[Double], generatedBy: Option[String] = None): A
+  def replace(full: A, part: B, generatedBy: Option[String] = None): A
 }
--- a/src/main/scala/scalismo/sampling/proposals/GaussianRandomWalkProposal.scala
+++ b/src/main/scala/scalismo/sampling/proposals/GaussianRandomWalkProposal.scala
@@ -1,31 +1,25 @@
 package scalismo.sampling.proposals
 
-import breeze.linalg.{DenseMatrix, DenseVector}
+import breeze.linalg.{DenseVector}
 import scalismo.sampling.{ProposalGenerator, SampleLens, TransitionProbability}
-import scalismo.statisticalmodel.MultivariateNormalDistribution
 
 /**
  * Classical Random Walk proposal, where the current state is perturbed using a step, which is generated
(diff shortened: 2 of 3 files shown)


Andreas-Forster:
Thank you for this work. I agree something like this is very useful. In principle, I have no strong objection to merging it.

But, I somehow do not like the usage of the SampleLense in the proposal's implementation. Already in the proposal, you then have to be aware, that you might have a bigger type of parameters and you always have to understand the SampleLense too if you want to write or use a proposal. Couldn't we have the proposals implemented on a concrete sample type and afterward use something as the following snippet for supporting other parameter constellations? I think this would improve the reuse of already existing code and even the SampleLense would fit this as a class containing the extract and emplace function given A and DenseVector[Double] as the type for PART.

```scala
import scalismo.sampling.{ProposalGenerator, TransitionProbability}

class PartialProposal[A, PART](
  val proposal: ProposalGenerator[PART] with TransitionProbability[PART],
  val extract: A => PART,
  val empalce: (A, PART) => A
) extends ProposalGenerator[A]
    with TransitionProbability[A] {
  override def propose(current: A): A = {
    empalce(current, proposal.propose(extract(current)))
  }

  override def logTransitionProbability(from: A, to: A): Double = {
    proposal.logTransitionProbability(extract(from), extract(to))
  }
}

object PartialProposal {
  object implicits {

    implicit class PartialProposer[PART](
      proposal: ProposalGenerator[PART] with TransitionProbability[PART]
    ) {
      def toSampleLenseProposal[A](extract: A => PART, empalce: (A, PART) => A) =
        new PartialProposal[A, PART](proposal, extract, empalce)
    }

  }
}
```

marcelluethi:
I see your point. It is really not ideal that the proposal generator has to be aware of the lense. It would make the code more clear if we could define it on a concrete sample instead. I also like the solution with the partial proposal (maybe with the exception of the implicit conversion). 

The main problem I see is that to follow this path we will need to be able to define what a full proposal is. This seems very much application specific. In an application where we fit a shape model to 2 fragments of a bone we will need different parameters than when we do 2D to 3D reconstruction and again different parameters are used when we use an intensity model. 

An alternative design to the sample lense would maybe be to drop the type parameter of the proposal generators completely. Instead we could implement all proposals on generic DenseVectors, which are then bestowed with meaning with a "SampleAdapter".

marcelluethi:
Unless we have a concrete sample type (which I argued above that I don't see how to construct) we need at some point to convert something to a dense vector. This was the original motivation of introducing the concept of the lense. I currently also don't see how generalizing the lense to a partial type would help here. Can you elaborate? My current thinking is, that if we have the lense, we would not need a partial proposal, as different lenses can be applied to the proposal generators, and in this way update part of the sample. 

Regarding the number of parameters: I introduced it for efficiency reasons. While the solution you propose works, it constructs a multivariate normal distribution every time a sample is generated. There is quite a bit of work involved in constructing a multivariate normal distribution, as for example computing the inverse covariance matrix. As this information needs to be known anyway for constructing the lense, I find it useful to expose it.

The last point you mentioned was the naming. My personal experience is that it makes debugging a lot easier if every sample is named. Having it as part of the interface would make it explicit that things should be named. If you really don't want to name a sample, you could still just ignore this parameter. In contrast, when it is a separate trait, you need to do extra work to name things. As I find that naming samples is always a good idea, I would opt for the leaving it there.  What is your experience with scalismo-faces? Do you usually name things or do you have another/better mechanism for debugging which we could adopt instead in Scalismo?

marcelluethi:
I just updated the PR. The SampleLens now takes two type Parameters, ```A```  and ```B``` for the sample and the part respectively. I also removed the numberOfParameters and updated the GaussianRandomWalkProposal accordingly. 

I have not  (yet?) removed the lense from the ```GaussianRandomWalkProposal```, as I am still not convinced that we ever need it without. It seems even to be the case that all the use-cases we have in Scalismo-faces could benefit from the lens. If we really ever want to sample a DenseVector directly, a trivial Identity-Lens would do it without introducing any overhead. 

From my side the uses cases I have in mind are covered with this implementation. However, I am open  to introduce the ProposalAdapter suggested above as well. 

Any thoughts on this update?

Andreas-Forster:
I think it looks now reasonable. I agree with the argument of the simple Identity-Lens for the GRWProposal.

I would be in favor of integrating the PartialProposal too. I used it already in several projects by copying. It would be nice to have the support in Scalismo directly. However, it does not necessarily have to be in this PR.

Andreas-Forster (approved):
I think the SampleLense and the GaussianRandomWalkProposal now make sense. Thank you for your work and the discussion.

Andreas-Forster:
Thank you for adding the PartialProposalAdapter. The adapter looks good to me.

PR #362, opened 2021-02-02
Improvement to GP marginal log likelihood accuracy

phoeft670:
switch to specialized logdet implementation that returns better results for small determinants


marcelluethi:
Thanks a lot for this PR. This is clearly the better way to compute the logdet.

PR #363, opened 2021-02-26
GP Regression with missing values

marcelluethi:
This commit achieves two things:
1) It changes the regression computations such that NaN observations are treated as missing observations and consequently removed from the computation. This makes it possible, that also a part of an observation can be specified (e.g. only the x component in a deformation vector). When no NaN values are present, the computations are exactly the same as now. 
2) The regression code was duplicated in the discreteLowRankGP and lowRankGP classes. This commit introduces a new class, which factors all the computations in one place.

This feature is becoming important now, as people start to work with more complicated output domain, e.g. building joint models of intensity and shape, or pose and shape. With this small change makes it becomes possible to use all the existing functionality, of scalismo also for this richer output domain. For example, we could condition on the intensity only, to get a posterior of the corresponding shape variations.

*This PR is meant as a discussion and should be thoroughly tested on such richer use cases before being merged*

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,8 +30,8 @@ lazy val root = (project in file("."))
       "org.scalanlp" %% "breeze" % "1.0",
       "org.scalanlp" %% "breeze-natives" % "1.0",
       "io.spray" %% "spray-json" % "1.3.5",
-      "ch.unibas.cs.gravis" % "scalismo-native-stub" % "4.0.0",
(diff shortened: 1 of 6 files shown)


marcelluethi:
I was just pushing an extension, where the use of partial data is only allowed when calling the regression function explicitly (as opposed to the posterior method). In addition, the strategy needs to be explicitly set when calling the function. 

@Andreas-Forster What do you think?

Andreas-Forster:
At first glance, I like the solution. It looks as it offers the desired new opportunities while still having a mechanism to not accidentally remove values.

I think we could give it a try like this.

PR #366, opened 2021-05-22
replaced TypeTag with enumeration in Scalar class

marcelluethi:
The use of a TypeTag to distinguish Scalartypes was always an ugly hack, which required the use of reflection. In this commit we replace all TypeTags with an explicit enumeration of the possible Scalartypes. The type is then passed along with the Scalar typeclass.

The original reason to make this change is the Scala 3 migration, where TypeTags seem to cause problem. It is, however, also more elegant and simplifies the type signature.


Andreas-Forster (approved):
I like the PR very much. Not only to get rid of the TypeTag is appealing, but also as it makes the transition towards Scala 3 most likely easier.

My only slight concern is that now we have two different types with the name `ScalarType`. Would it make sense to maybe change the name of the class derived from Enumeration in ImageIO.scala to ScalarVTKTypes?

marcelluethi:
Thanks for the review. You raise a good point, which I addressed in the latest commit. I renamed the class to ScalarDataType (which is still rather generic, but at least makes it clear that it is different from Scalar), moved it into a separate File within the io package and made it private. It is only used internally for the io methods and vtk interoperability and a user should never directly use it. 

I believe that in this way no confusion can arise.

PR #368, opened 2021-06-26
Two small improvements/bugfixes

marcelluethi:
This PR addresses two minor problems in v0.90.0. As both affect the API, we only change them in the upcoming 0.91 release. 

1. The internal model of an Active shape model should be a ```PointDistributionModel[_3D, TriangleMesh]``` and not a ```StatisticalMeshModel```
2. The method ```pointWithId``` on ```DiscreteField``` should return an iterator of ```(Point[D], PointId)``` instead of an iterator ```(Point[D], int)```.


Andreas-Forster (approved):
Thank you for the fixes. I think they are reasonable.

PR #372, opened 2021-08-24
fixed the log normalization for the logpdf.

### LINKED ISSUE #371: Error in LogPdf computation of MVND
The `logpdf` function in the `MultivariateNormalDistribution` fails to correctly compute the logpdf in cases where a large covariance matrix is defined with a small variance. 
In such a case the normFactor is not numerical stable. 

An example of this is a diagonal matrix with a std. dev < 1.0 and a rank > 50:
 
```scala
 val stdDev = 0.1
 val model = myMVND(DenseVector.zeros[Double](100), DenseMatrix.eye[Double](100)*stdDev*stdDev)
 model.logpdf(DenseVector.zeros[Double](100))
```
this gives -80.38 

With `scipy` in `python`, the output is instead 138.36

JonathanAellen:
Fixes #371 the log normalization for covariance values that are small and lead to numerical issues (->0.0) in higher dimensions (>50). also added two tests to check the numerical stability for diagonal and non-diagonal covs.


marcelluethi:
Thank you very much for the fix - and the tests.

PR #374, opened 2021-12-30
Make scalismo work with Scala 3

marcelluethi:
This PR makes scalismo crosscompile for scala 3.0 and scala 2.13.


madsendennis:
Seems good to me. Test passes locally for me and I've also checked that I can use the version for a project with both scala 3 and 2.13 main files.

marcelluethi:
@madsendennis Thanks for looking at it. In the meantime I fixed the formatting also found out what caused the tests to fail (sometimes). It was a problem with the initialization during object creation time. More precisely, I computed a matrix inverse to initialize a val in an object. As it is not determined when the values in the objects are created, this sometimes happened before netlib was properly initialized. Changing the corresponding val to a lazy val solved the problem.

Andreas-Forster (commented):
To me the changes looked reasonable and I was able to run and pass all tests for scala 3.0. However, when cross-compiling the tests with scala 2.13.6 it seems to produce errors on my machine.

It reports some unused imports, but later on, it states that some of the classes are not found.  I will try to soon find some time to look into that.

Andreas-Forster (approved):

PR #379, opened 2022-03-29
Added optional scale argument to gaussian kernel

marcelluethi:
To create a meaningful deformation model using a GP with a Gaussian Kernel, two parameters need to be specified: the standard deviation, which models the smoothness, and the scale, which determines the amplitude (size) of the modelled functions. So far only the smoothness could be specified when constructing the Gaussian kernel. The scale was specified by multiplying the resulting kernel with a factor $s^2$. This commit adds an optional parameter to the Constructor, which makes it possible to add the factor already during construction.

This makes the intention much clearer. Furthermore, the scale factor has the unit mm, whereas when multiplying, the square of the deformation needed to be passed.

### DIFF
--- a/src/main/scala/scalismo/kernels/StandardKernels.scala
+++ b/src/main/scala/scalismo/kernels/StandardKernels.scala
@@ -22,27 +22,34 @@ import scalismo.numerics.BSpline
 import scalismo.transformations.Transformation
 import scalismo.utils.Memoize
 
-case class GaussianKernel[D](sigma: Double) extends PDKernel[D] {
+/**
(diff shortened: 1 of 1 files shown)


Andreas-Forster:
I like the better interface for constructing kernels and that the parameter is not the squared one. However, this is also a bit of a drawback, as the parameter when scaling the kernel after constructing it needs to be squared to generate the same effect. This might cause some confusion. I guess we can not do much about it.

In summary, I like the PR and I am in favor of adding this improved interface. Thank you for the work.

Andreas-Forster (approved):

PR #380, opened 2022-04-07
Added check in gradient computation that avoids nan at extremum

marcelluethi:


src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -116,8 +116,10 @@ class TriangleMesh3DOperations(private val mesh: TriangleMesh[_3D]) {
 
     def grad(pt: Point[_3D]) = {
       val closestPt = closestPoint(pt).point
-      val grad = EuclideanVector(pt(0) - closestPt(0), pt(1) - closestPt(1), pt(2) - closestPt(2))
-      grad * (1.0 / grad.norm)
+
+      val grad = closestPt - pt
+      val gradNorm = grad.norm
+      if (gradNorm < 1e-10) EuclideanVector.zeros[_3D] else grad * (1.0 / gradNorm)
```
Andreas-Forster, line 122: I agree that we should guard here the division by zero. I have not thought, if this threshold makes sense, as usually the norm is in a similar range than the values of the vector itself. So it should be fine as long as it is not zero. I agree that we usually assume something in the range of 10+2 to 10-2 maybe hence the threshold should be fine.

src/main/scala/scalismo/mesh/MeshOperations.scala
```diff
@@ -116,8 +116,10 @@ class TriangleMesh3DOperations(private val mesh: TriangleMesh[_3D]) {
 
     def grad(pt: Point[_3D]) = {
       val closestPt = closestPoint(pt).point
-      val grad = EuclideanVector(pt(0) - closestPt(0), pt(1) - closestPt(1), pt(2) - closestPt(2))
-      grad * (1.0 / grad.norm)
+
+      val grad = closestPt - pt
```
Andreas-Forster, line 120: Doesn't that change the direction of the gradient? before it was pt - closestPoint no?

Andreas-Forster (changes requested):
I agree that it would be good to guard this case. I have however a question about the direction of the gradient. Once discussed I am fine with merging.

src/main/scala/scalismo/mesh/MeshOperations.scala
marcelluethi: nice catch. Thanks for pointing this out.

PR #381, opened 2022-04-07
added new method to compute center of mass

marcelluethi:
The center of mass of a point set is needed in many optimization and sampling algorithm. In order to avoid having to compute it all the time, I propose to introduce a convenience method on ```PointSet```, which  returns the center of mass


Andreas-Forster (approved):
Happy to see this coming to the library. So often already coded it but never made a PR 😬. Thank you for doing it 👍

PR #382, opened 2022-04-08
updated marginal likelihood

JonathanAellen:
Updated the marginal likelihood for the various GP classes.


marcelluethi (approved):
This looks good to me.

marcelluethi:
Thanks for adding this functionality. It is great that we have these methods now available in all 4 variants of the GPs.

PR #383, opened 2022-04-20
fixed bug in relabel method of MHProposalGenerator

marcelluethi:
Relabeling a sample would not return a new proposal, but always the current one. This commit fixes this problem.


Andreas-Forster (approved):
This bug fix makes absolute sense.

Thank you for continuously improving things.

PR #384, opened 2022-04-20
added mhcombinedproposal

marcelluethi:
The MHCombinedProposal mimicks the CombinedProposal, but is adapted to the new MHSample type.


Andreas-Forster (approved):
I think it makes sense to have such a proposal and the implementation looks reasonable on first sight.

Thank you for the contribution.

PR #386, opened 2022-04-25
fixed wrong computation of logTransitionProbability (fixes #385)

### LINKED ISSUE #385: Shouldn't the log transition probability depend on the induced change?
Not sure if I completely oversee something, but shouldn't the log transition probability depend on the change that the sample induced?
https://github.com/unibas-gravis/scalismo/blob/1d7137171101c84ce9912bb30eb82d2e149fc525/src/main/scala/scalismo/sampling/proposals/GaussianRandomWalkProposal.scala#L45

I would expect that it depends on something like (to.parameters-from.parameters) as larger changes should usually have a smaller transition probability than smaller once.

marcelluethi:
The logTransitationProbability did only return the normalization
term of the gaussian distribution rather than the actual log
probability.


Andreas-Forster (approved):
Thank you for the fix for the missing calculation.

PR #387, opened 2022-05-14
use jhdf instead of the hdf5 native library for reading hdf5 files

marcelluethi:
There exist a number of scurity vulnerabilities in the older hdf5 implementation
that scalismo is using (see https://github.com/unibas-gravis/scalismo-native/issues/6). 

This PR changes all reading routines for hdf5 to use jhdf, which is a pure java
implementation. Writing is still done using hdf5. The implementation is
done such that much of the original structure is preserved. Once we get
completely rid of the native hdf5 libraries, a refactoring should be
done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ lib/libvtk*
 /project/project
 /.idea/
 /*.iml
+null/
+.bloop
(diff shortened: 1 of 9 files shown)


Ghazi-Bouabene (approved):
Thank you very much for these changes.

I have tested loading a legacy model (previously written with ncsa hdf), generated random meshes and saved to .ply.

Did the same using scalismo 0.90 (and hence ncsa HDF for reading).

The generated meshes are identical (m5sum of the files is the same).

Andreas-Forster (approved):
Thank you a lot for the work. The most changes look reasonable to me. As it was already tested, I did not test it too.

Just one little question, do we need now these ObjectToArrayCasts for multiple dimensions? Or is this to make it more future proved?

PR #388, opened 2022-06-07
removed private modifier in UnstructuredPointsXD case classes

marcelluethi:
The constructors were made private in an early version of Scalismo,
where the philosophy was that all classes should be constructed using
a generic dimension argument. This has since been changed, but here
we seem to have forgotten to update the code.

 Please enter the commit message for your changes. Lines starting


madsendennis (approved):
Looks good to me.

PR #389, opened 2022-07-01
numerical stability in low rank marginal likelihood

JonathanAellen:
Previously the calculation was unstable when used in models with very small eigenvalues (<1e-7).

The proposed solution filters out these eigenvalues. Otherwise the same algorithm is used. That means a model with the following eigenvalues/vectors:
`
evec = [v1,v2,v3,v4] 
eval = [1,1,1,1e-10]
`
is treated as the following model
`
evec = [v1,v2,v3]
eval = [1,1,1]
`

### DIFF
--- a/src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
+++ b/src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
@@ -218,9 +218,11 @@ case class DiscreteLowRankGaussianProcess[D: NDSpace, DDomain[DD] <: DiscreteDom
   override def marginalLikelihood(td: IndexedSeq[(PointId, Value, MultivariateNormalDistribution)]): Double = {
     require(td.nonEmpty, "provide observations to calculate the marginal likelihood")
     val matIds = td.flatMap(t => t._1.id * outputDim until t._1.id * outputDim + outputDim)
-    val U = DenseMatrix.tabulate[Double](matIds.length, rank)((x, y) => basisMatrix(matIds(x), y))
+    val r = variance.data.count(_ > 1e-7)
+    val S = if (r == rank) variance else DenseVector(variance.data.take(r))
+    val U = DenseMatrix.tabulate[Double](matIds.length, r)((x, y) => basisMatrix(matIds(x), y))
     val y = DenseVector(td.flatMap(t => (vectorizer.vectorize(t._2) - vectorizer.vectorize(mean(t._1))).toArray): _*)
-    LowRankGaussianProcess.marginalLikelihoodComputation(U, variance, y, td.map(_._3))
+    LowRankGaussianProcess.marginalLikelihoodComputation(U, S, y, td.map(_._3))
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
```diff
@@ -218,9 +218,11 @@ case class DiscreteLowRankGaussianProcess[D: NDSpace, DDomain[DD] <: DiscreteDom
   override def marginalLikelihood(td: IndexedSeq[(PointId, Value, MultivariateNormalDistribution)]): Double = {
     require(td.nonEmpty, "provide observations to calculate the marginal likelihood")
     val matIds = td.flatMap(t => t._1.id * outputDim until t._1.id * outputDim + outputDim)
-    val U = DenseMatrix.tabulate[Double](matIds.length, rank)((x, y) => basisMatrix(matIds(x), y))
+    val r = variance.data.count(_ > 1e-7)
+    val S = if (r == rank) variance else DenseVector(variance.data.take(r))
```
marcelluethi, line 222: wouldn't it be equivalent (but simpler) just write S := variance(0 until r)

src/main/scala/scalismo/statisticalmodel/LowRankGaussianProcess.scala
```diff
@@ -200,7 +200,7 @@ class LowRankGaussianProcess[D: NDSpace, Value](mean: Field[D, Value], val klBas
       val data = vectorsSeq.flatMap(value => vectorizer.vectorize(value).toArray).toArray
       val vector = new DenseMatrix(td.length * this.outputDim, 1, data)
       (lambda, vector)
-    }).unzip
+    }).filter(t => t._1 > 1e-7).unzip
```
marcelluethi, line 203: Maybe 1e-7 is not small enough as a threshold? See my comment above.

marcelluethi (changes requested):
Thanks for the PR. I think this change makes sense and we should merge it. 
I just have a few comments/nitpicks.

src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
```diff
@@ -218,9 +218,11 @@ case class DiscreteLowRankGaussianProcess[D: NDSpace, DDomain[DD] <: DiscreteDom
   override def marginalLikelihood(td: IndexedSeq[(PointId, Value, MultivariateNormalDistribution)]): Double = {
     require(td.nonEmpty, "provide observations to calculate the marginal likelihood")
     val matIds = td.flatMap(t => t._1.id * outputDim until t._1.id * outputDim + outputDim)
-    val U = DenseMatrix.tabulate[Double](matIds.length, rank)((x, y) => basisMatrix(matIds(x), y))
+    val r = variance.data.count(_ > 1e-7)
```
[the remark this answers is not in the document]
JonathanAellen: Sounds good. I think a 1e-10 is enough as for higher point count this is already modelling a very small deformation. This could still prove to be an issue for small shapes (contained in some eps-radius) with only few points. But for these a scaling change seems anyway a reasonable thing to do.

marcelluethi (approved):

Issue #390, opened 2022-07-05
Error reading .h5 file


Ghazi-Bouabene:
Using version 0.91.1 I run into this exception when reading a legacy .h5 file

Exception in thread "main" java.lang.ClassCastException: class java.lang.String cannot be cast to class [Ljava.lang.String; (java.lang.String and [Ljava.lang.String; are in module java.base of loader 'bootstrap')
	at scalismo.io.HDF5Reader.$anonfun$readStringAttribute$1(HDF5Utils.scala:57)
	at scala.util.Try$.apply(Try.scala:210)
	at scalismo.io.HDF5Reader.readStringAttribute(HDF5Utils.scala:60)
	at example.ExampleApp$.main(ExampleApp.scala:147)
	at example.ExampleApp.main(ExampleApp.scala)

The representer name in the file is a pretty standard one (itkStandardMeshRepresenter).

My guess is that this is due to casting a String to an Array[String] here :
https://github.com/unibas-gravis/scalismo/blob/18c327d8656c7f3f35e458b7ff42318222bee68a/src/main/scala/scalismo/io/HDF5Utils.scala#L59

I have tried a version where I cast to String instead and that works well. Is there a reason why we cast to Array here? (and also for other attribute types)

marcelluethi:
Good to know that it only happens in legacy files. 

Yes, we should fix this and also the subsequent method, where we have the same bug for attributes of type int. 
Once this is fixed, I can release a version 0.91.2. Do you want to do a PR?

PR #391, opened 2022-07-06
Bug fix reading legacy h5

Ghazi-Bouabene:
Fixes issue #390 

Had to add another cast to the one in the discussion since NDArray reading seemed also to fail.


marcelluethi (approved):
Thanks for fixing this issue. It looks good to me.

Ghazi-Bouabene:
There should be no further issues by now.

Issue #367, opened 2021-06-16
Non-deterministic behaviour of gradient registration


thogerig:
The following snippet for testing the numerical consistency of the registration output lead to non-deterministic behaviour since not all test-runs gave exactly the same output.
```
...
    val iterations = (for(i <- 0 until 10) yield {

      val regIt = Registration(
        metric,
        L2Regularizer(transformationSpace),
        regularizationWeight = 0.0,
        LBFGSOptimizer(maxNumberOfIterations = 300)
      ).iterator(DenseVector.zeros[Double](transformationSpace.numberOfParameters))
      val regItPrinting = for (it <- regIt) yield {
        it.value
      }
      regItPrinting.toIndexedSeq.sum
    }).toIndexedSeq


    val firstResult = iterations.head
    for (i <- 0 until iterations.length) {
      iterations(i) should be(firstResult)
    }
```
The reason seems to be an issue of the parallel calculation of the sum in the MeanPointwiseLossMetric.scala in line 81. Changing there the line to the following by adding a toIndexedSeq, the test passes and all runs are consistent:
```
new ParVector(samples.toVector).map { case (pt, _) => metricValue(pt).getOrElse(0.0) }.toIndexedSeq.sum / samples.size
``` #
I will make a PR with a test and a fix.

thogerig:
Interestingly, I did now run the consistency test on the current develop and the issue seems to be fixed.

Andreas-Forster:
Unfortunately, this is not yet fixed. I tested on the recent develop branch at v0.91-RC3. The MeanPointwiseLossMetric uses the ParVector and therefore has some numerical inaccuracies. I used the snippet below to check it:

```scala
import breeze.linalg.{DenseMatrix, DenseVector}
import scalismo.common.interpolation.TriangleMeshInterpolator3D
import scalismo.common.{DifferentiableField, Domain, EuclideanSpace, EuclideanSpace3D, Field, PointId, Scalar}
import scalismo.geometry.{EuclideanVector, NDSpace, Point, _3D}
import scalismo.kernels.{DiagonalKernel3D, GaussianKernel3D, MatrixValuedPDKernel}
import scalismo.mesh.{TriangleCell, TriangleList, TriangleMesh, TriangleMesh3D}
import scalismo.numerics.{FixedPointsUniformMeshSampler3D, Sampler}
import scalismo.registration.RegistrationMetric.ValueAndDerivative
[...]

PR #392, opened 2022-08-22
Bug fix: avoid using caching in SurfaceSpatialIndex as this results i…

Ghazi-Bouabene:
…n different values for the same point when ran in parallel. Fixes issue #367

### DIFF
--- a/src/main/scala/scalismo/mesh/boundingSpheres/SurfaceSpatialIndex.scala
+++ b/src/main/scala/scalismo/mesh/boundingSpheres/SurfaceSpatialIndex.scala
@@ -188,7 +188,7 @@ private[mesh] class TriangleMesh3DSpatialIndex(private val bs: BoundingSphere,
     val p = point.toVector
 
     // last triangle might be a good candidate
-    val result = BSDistance.toTriangle(point.toVector, triangles(lastIdx.get().idx))
+    val result = BSDistance.toTriangle(point.toVector, triangles(0))
     updateCP(res.get(), result)
 
     // search for true candidate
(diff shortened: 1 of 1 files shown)


src/main/scala/scalismo/mesh/boundingSpheres/SurfaceSpatialIndex.scala
```diff
@@ -188,7 +188,7 @@ private[mesh] class TriangleMesh3DSpatialIndex(private val bs: BoundingSphere,
     val p = point.toVector
 
     // last triangle might be a good candidate
```
marcelluethi, line 190: I guess this comment does not fit the code anymore and should be updated.

marcelluethi (commented):
Thanks for the PR. I cannot assess the implications of this change and need to defer to @Andreas-Forster. If it is fine for Andreas, it is okay for me.

Andreas-Forster (approved):
I agree with the PR that it fixes the issue we have right now. While it might be that it gets slower, it is definitively more deterministic.

When we need the last bit of performance we would have to check if there is a way to fix this issue, by improving the algorithm. But this does not have priority.

marcelluethi:
@Andreas-Forster I realized only now that I don't understand the issue very well. Could you quickly comment on why taking the last points as a starting point makes it non-deterministic, while taking the head does not? Is this also due to numerical instabilities or is there a bug hidden somewhere that we should fix?

Andreas-Forster:
I did not debug it in detail. But I guess the tree traversal could be different for a point when you have a different initialization there. Just the order of visiting the nodes. I guess that sometimes then are two points found that have the (nearly) same distance and hence the order matters. Then starting always at the same point, makes the tree traversal agnostic of the last found point.

PR #393, opened 2023-02-12
replace dependency on spray-json with ujson

marcelluethi:
Replace the dependency on spray-json with that of upickle/ujson. 

[upickle/ujson](https://github.com/com-lihaoyi/upickle) is a simple, dependency-free json parser. It has been chosen as the basis for the upcoming [scala toolkit](https://virtuslab.com/blog/scala-toolkit-makes-scala-powerful-straight-out-of-the-box/) and hence it is guaranteed that it is actively maintained also in the future.  As an additional advantage, it supports binary formats, such as messagepack, in case we ever need it for a new binary format for writing statistical shape models.


Andreas-Forster (approved):
I think this switch is reasonable. Thank you for implementing it.

PR #395, opened 2023-02-26
Statistical Model IO to hdf5-json

marcelluethi:
This PR introduces a major change of how statistical models are written in Scalismo. 
Writing no longer supports hdf5 directly, but only the related hdf.json format, which in 
turn can be converted to hdf5 using [tools provided by the hdfgroup](https://github.com/HDFGroup/hdf5-json). 
For this purpose, a separate scalismo-hdf5-json library is added as a dependency. 

The PR also moves all the statistical model io files into an own directory. 
Finally, instead of using strings to represent path in HDF files, a new class HDFPath is used, 
which adds typesafety to the method signatures.


src/main/scala/scalismo/io/MeshIO.scala
```diff
@@ -648,22 +648,22 @@ object MeshIO {
...
-        h5file.close()
-      }
-    } yield {
-      TriangleMesh3D(UnstructuredPoints(NDArrayToPointSeq(vertArray).toIndexedSeq),
-                     TriangleList(NDArrayToCellSeq(cellArray)))
-    }
-
-    maybeSurface
-  }
+//  def readHDF5(file: File): Try[TriangleMesh[_3D]] = {
```
Andreas-Forster, line 651: Could we remove this commented out code?

build.sbt
```diff
@@ -63,8 +63,8 @@ lazy val root = (project in file("."))
       ("org.scalanlp" %% "breeze" % "2.1.0"),
       ("org.scalanlp" %% "breeze-natives" % "2.1.0"),
       "ch.unibas.cs.gravis" % "scalismo-niftijiojar" % "0.1.0",
-      "com.lihaoyi" %% "upickle" % "2.0.0",
-      "ch.unibas.cs.gravis" % "hdf5javanatives" % "0.1.0",
+      //"com.lihaoyi" %% "upickle" % "2.0.0",
+      ("ch.unibas.cs.gravis" %% "scalismo-hdf5-json" % "0.1.0-SNAPSHOT").cross(CrossVersion.for2_13Use3),
```
Andreas-Forster, line 67: I would be in favor of having here a non SNAPSHOT version, or is there a specific reason for it?

Andreas-Forster (commented):
Thank you for this huge PR. I'm really looking forward to seeing it merged.

I haven't reviewed it in full detail, but have two questions.

build.sbt
marcelluethi: This will be a non-snapshot version in the final version. But I think we should plan a phase for testing the new io classes with real use cases before the release. As scalismo-hdf5-json is brand new, I expect quite that we will encounter some issues. 

I propose the following strategy:

- Test everything together with Scalismo-ui using the SNAPSHOT release
- Once we have initial tests done, create RCs for all libraries and test those
- Once the RCs are tested for some weeks, we release the final versions.

Andreas-Forster: That makes sense.

PR #396, opened 2023-02-26
drop crossScalaVersion 2.13

marcelluethi:
So far we compiled scalismo for both scala 3 and scala 2.13. This restricts the syntax to a subset of the language and forces the downstream libraries to be available as 2.13. As Scala 2.13 is able to consume libraries written in Scala 3.x, I would argue that we can safely drop scala 2.13 and focus instead on writing clean scala 3.x code, making use of the many language improvements.

### DIFF
--- a/src/main/scala/scalismo/io/StatisticalModelIO.scala
+++ b/src/main/scala/scalismo/io/StatisticalModelIO.scala
@@ -15,7 +15,9 @@
  */
 package scalismo.io
 
-import java.io._
+import scalismo.common.interpolation.NearestNeighborInterpolator3D
+
+import java.io.*
 import scalismo.common.{Scalar, UnstructuredPointsDomain, Vectorizer}
 import scalismo.geometry.{_1D, _2D, _3D, EuclideanVector}
 import scalismo.image.DiscreteImageDomain
@@ -240,7 +242,7 @@ object StatisticalModelIO {
    * @return statisticalMeshModel
    */
   private def modelConverterToMeshModel(model: PointDistributionModel[_3D, TriangleMesh]): StatisticalMeshModel = {
-    StatisticalMeshModel(model.reference, model.gp)
+    StatisticalMeshModel(model.reference, model.gp.interpolate(NearestNeighborInterpolator3D()))
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/scala/scalismo/registration/MutualInformationMetric.scala
```diff
@@ -208,7 +208,7 @@ case class MutualInformationMetric[D: NDSpace, A: Scalar](fixedImage: Field[D, A
       val derivsAtValidPoints = derivsForPoints.collect { case Some(v) => v }
       val deriv = derivsAtValidPoints.foldLeft(zeroVec)((acc, v) => acc + v)
 
-      val scaledDeriv = (1.0 / (derivsAtValidPoints.size * binSizeMovingImage)) * deriv
+      val scaledDeriv = deriv * (1.0 / (derivsAtValidPoints.size * binSizeMovingImage))
```
Andreas-Forster, line 211: 1.0 /  is no longer needed. should we keep this for clarity?

src/main/scala/scalismo/io/StatisticalModelIO.scala
```diff
@@ -15,7 +15,9 @@
  */
 package scalismo.io
 
-import java.io._
+import scalismo.common.interpolation.NearestNeighborInterpolator3D
+
+import java.io.*
```
Andreas-Forster, line 20: Can we make scalafmt sorting the includes?

Andreas-Forster (approved):
I think the idea of dropping 2.13 when the library stays compatible with 2.13 makes total sense.

Few comments, but totally fine with merging as is.

src/main/scala/scalismo/registration/MutualInformationMetric.scala
marcelluethi: Not sure that I understand. It should still be there, but we can only put it right of the vector. For some reason with scala 3.2 the multiplication with the double on the lhs of the * operator is causing a compile time error. This is why I had to make this change everywehre.

src/main/scala/scalismo/io/StatisticalModelIO.scala
marcelluethi: I will look into it and make a separate PR with a new scalafmt. Once we got rid of scala 2.13 we have a lot more freedom to clean up things and enforce new formatting rules.

PR #398, opened 2023-02-27
add explicit constructors for 1D/2D/3D DiscreteLowRankGPs

marcelluethi:
As a convention, all the dimension dependent objects should be created using an explicit dimension. For DiscreteLowRankGPs these constructors were missing. This commit adds these constructors. At the same time, DiscreteLowRankGP is changed from a case class to a regular class, which makes more sense at it is clearly not a class that holds only data.


Andreas-Forster (approved):
Thank you for the PR. I think these changes makes totally sense.

PR #401, opened 2023-03-24
add given instances to make it easier to define a random or fixed seed

marcelluethi:
To use a fixed seed for the random number generator, an explicit given instance had to be created by the user. Following the lead of breeze, this commit introduces instead a given instance for fixed and variable seed, which can be imported by the user. Scala 3 will produce a neat error message, if the import is missing.


marcelluethi:
@Andreas-Forster I am aware that this PR does not pass scalafmt checks. The reason is that we need to upgrade to a more recent version of scalafmt first, which understands scala 3 syntax. I will do this in a separate PR.

Andreas-Forster (approved):
Thx - the message is indeed very nice.

PR #405, opened 2023-06-14
bump scala to version 3.3 and squash some warnings

marcelluethi:
This PR is the first one in the sequence of several PRs to complete scalismo version 0.92

PR #406, opened 2023-06-16
Add 3D constructor to ScalarMeshField

marcelluethi:
An explicit constructor for the 3D case was missing, which should be available for all geometric objects. 
This PR adds the constructor and deprecates the generic `ScalarMeshField` constructor. 

Please merge only after #405 is merged.


Andreas-Forster (approved):
Thank you for adding this.

PR #407, opened 2023-06-29
add method to join two meshes together

marcelluethi:
A common use case in scalismo is to build a model from individual parts (such as the ulna and radius bone). To be able to build a joint model, we need a reference mesh, that is comprised of both of them. So far scalismo did not offer a simple way to join two meshes together. This PR adds this feature for 3D triangle meshes and 3D tetrahedral meshes.

### DIFF
--- a/src/main/scala/scalismo/mesh/MeshOperations.scala
+++ b/src/main/scala/scalismo/mesh/MeshOperations.scala
@@ -15,7 +15,7 @@
  */
 package scalismo.mesh
 
-import scalismo.common.{DifferentiableField, EuclideanSpace, Field, PointId}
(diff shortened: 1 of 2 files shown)


Andreas-Forster (approved):
Thank you for the PR. I think this method is really useful. For sure, already a lot of Scalismo-based code has an implementation for it, hopefully always correct.

A minor point is that the tests are OK, but not super sensitive.

marcelluethi:
You are right. I just pushed a (hopefully) improved version of the tests, which explicitly create a ground truth mesh to compare to.

Andreas-Forster:
Yes, I think this does indeed make the test more selective. Thx

PR #408, opened 2023-06-29
fix read/write trianglemesh tests (fixes #404)

### LINKED ISSUE #404: Bug in MeshIO tests - not testing for different file types
The following function is called twice, but the passed parameter (the different extension) is never used.

https://github.com/unibas-gravis/scalismo/blob/21232cbab1049f0adcfaf8b5dafcea0ee771d4f1/src/test/scala/scalismo/io/MeshIOTests.scala#L40

marcelluethi:


Andreas-Forster (approved):
Thanks a lot for fixing it.

PR #409, opened 2023-06-30
add samplePoints method to sampler

marcelluethi:
This PR proposes to add a convenience function,  `samplePoints` to a sampler. 
So far only the `sample` method was available, which returns a tuple of points and their probability. This is useful for internal algorithms, but often not what the user wants to see. As we cannot remove the public `sample` method, a new method `samplePoints` is added, which simply throws away the probabilities.


Andreas-Forster (approved):
Thank you for adding this new functionality.

Issue #413, opened 2023-12-26
Slow triangular mesh boundary calculation for large meshes


madsendennis:
For large meshes (+100K vertices), the current calculation `TriangularMeshBoundaryPredicates` is slow. 
E.g. a 400K vertice mesh can end up taking +30 seconds.

From debugging, it seems that the slow computation is from a continuous update of a `CSCMatrix[Boolean]`:
```
    triangles.foreach { triangle =>
      edgesOfATriangle.foreach { edge =>
        val v1 = triangle.pointIds(edge._1).id
        val v2 = triangle.pointIds(edge._2).id
        edgeOnBorderTmp(v1, v2) = !edgeOnBorderTmp(v1, v2)
        edgeOnBorderTmp(v2, v1) = !edgeOnBorderTmp(v2, v1)
      }
    }
```

marcelluethi:
This issue has been resolve (PR #414 ) and published as v0.92.1

PR #414, opened 2023-12-26
Fix/slow boundary calculation for large meshes

madsendennis:
Fixes a speed issue with calculation boundary points on large triangle meshes.
 - add intermediate edge set
 - add boundary test on face mesh
This addresses issue https://github.com/unibas-gravis/scalismo/issues/413

### DIFF
--- a/src/main/scala/scalismo/mesh/MeshBoundaryPredicates.scala
+++ b/src/main/scala/scalismo/mesh/MeshBoundaryPredicates.scala
@@ -170,21 +170,29 @@ object MeshBoundaryPredicates {
       }
     }
     val edgeOnBorderTmp = edgeOnBorderBuilder.result()
-
     // edges from only a single triangle lie on the border,
     // edges contained in two triangles are not on a border
+    var edgeOnBorderSet: Set[(Int, Int)] = Set.empty
     triangles.foreach { triangle =>
       edgesOfATriangle.foreach { edge =>
         val v1 = triangle.pointIds(edge._1).id
         val v2 = triangle.pointIds(edge._2).id
-        edgeOnBorderTmp(v1, v2) = !edgeOnBorderTmp(v1, v2)
-        edgeOnBorderTmp(v2, v1) = !edgeOnBorderTmp(v2, v1)
+
+        if (edgeOnBorderSet.contains((v1, v2)) || edgeOnBorderSet.contains((v2, v1))) {
+          edgeOnBorderSet -= ((v1, v2))
+          edgeOnBorderSet -= ((v2, v1))
+        } else {
+          edgeOnBorderSet += ((v1, v2))
+        }
       }
+    }
+    edgeOnBorderSet.foreach{case(v1, v2) =>
+      edgeOnBorderTmp(v1, v2) = true
+      edgeOnBorderTmp(v2, v1) = true
     }
 
     // optimization : we do not need values corresponding to false in the sparse representation
     val edgeOnBorder = reduceCSCMatrixBooleansToTrueEntries(edgeOnBorderTmp)
-
     // points at one end of a border edge are also on the border,
     // triangles with one side on the border are also on the border
     triangles.zipWithIndex.foreach { t =>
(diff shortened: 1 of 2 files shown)


madsendennis:
This minor "fix" is much faster for large meshes. E.g. a 400K vertices mesh on my machine, that took +40 seconds with the previous version now takes 1 second. 
I know that introducing an intermediate set variable seems messy, so I'm open to alternatives.

src/main/scala/scalismo/mesh/MeshBoundaryPredicates.scala
```diff
@@ -170,21 +170,29 @@ object MeshBoundaryPredicates {
...
+        }
       }
     }
+    edgeOnBorderSet.foreach{case(v1, v2) =>
+      edgeOnBorderTmp(v1, v2) = true
+      edgeOnBorderTmp(v2, v1) = true
+    }
 
     // optimization : we do not need values corresponding to false in the sparse representation
     val edgeOnBorder = reduceCSCMatrixBooleansToTrueEntries(edgeOnBorderTmp)
```
Andreas-Forster, line 195: Thank you for looking into it and propose the improvement. I think with the proposed changes, we can even modify the code more. Did not test it, but I would propose something like the following for the function:

```scala
  /**
   * Build boundary index for triangle mesh.
   *
   * @param mesh
   *   Incoming triangle mesh.
   * @return
   *   Boundary that can be queried for by index for points, edges, and triangles.
   */
  def apply[D](mesh: TriangleMesh[D]): TriangularMeshBoundaryPredicates = {

    val points = mesh.pointSet.points.toIndexedSeq
    val nPts = points.length

    val triangles = mesh.triangulation.triangles
    val nTriangles = triangles.length
    val edgesOfATriangle = Seq((0, 1), (1, 2), (2, 0))

    val pointOnBorder = new Array[Boolean](nPts)
    val triangleOnBorder = new Array[Boolean](nTriangles)

    // edges from only a single triangle lie on the border,
    // edges contained in two triangles are not on a border
    var edgeOnBorderSet: Set[(Int, Int)] = Set.empty
    triangles.foreach { triangle =>
      edgesOfATriangle.foreach { edge =>
        val v1 = triangle.pointIds(edge._1).id
        val v2 = triangle.pointIds(edge._2).id

        if (edgeOnBorderSet.contains((v1, v2)) || edgeOnBorderSet.contains((v2, v1))) {
          edgeOnBorderSet -= ((v1, v2))
          edgeOnBorderSet -= ((v2, v1))
        } else {
          edgeOnBorderSet += ((v1, v2))
        }
      }
    }
    
    val edgeOnBorderBuilder = new CSCMatrix.Builder[Boolean](nPts, nPts)
    edgeOnBorderSet.foreach{case(v1, v2) =>
      edgeOnBorderBuilder.add(v1, v2, true)
      edgeOnBorderBuilder.add(v2, v1, true)
    }
    val edgeOnBorder = edgeOnBorderBuilder.result()
    
    // points at one end of a border edge are also on the border,
    // triangles with one side on the border are also on the border
    triangles.zipWithIndex.foreach { t =>
      val triangle = t._1
      val tIdx = t._2
      edgesOfATriangle.foreach { edge =>
        val v1 = triangle.pointIds(edge._1).id
        val v2 = triangle.pointIds(edge._2).id
        if (edgeOnBorder(v1, v2)) {
          pointOnBorder(v1) = true
          pointOnBorder(v2) = true
          triangleOnBorder(tIdx) = true
        }
      }
    }

    new BoundaryOfATriangleMeshPredicates(pointOnBorder.toIndexedSeq, edgeOnBorder, triangleOnBorder.toIndexedSeq)
  }
```

madsendennis: Excellent suggestion. That indeed makes much more sense. I've tested and pushed the suggestion.

Andreas-Forster (approved):
Thank you for your work. To me, it looks now ready to merge.

marcelluethi (approved):
Thanks a lot, Dennis, for this improvement. 
It looks good to me.

Issue #415, opened 2024-01-07
Remove VTK dependency


madsendennis:
Scalismo currently relies on the VTK library for many computations. This dependency is the main reason that Scalismo does not easily run in a container environment. 
To make that work, VTK itself also needs to be installed in the container (which makes the containers a lot larger).

The VTK library is also often far from fast - probably due to converting to/from the vtk format. 

I would therefore suggest to eventually have the VTK dependency removed from Scalismo. 

This would have implications on:
- Mesh decimation
- Mesh I/O
- Tetrahedral mesh

And a lot of:
- Image I/O and operations

Maybe some of the things are not needed to be supported in the future, which is why I opened this ticket as a discussion as well to hear more view points on this "problem".

marcelluethi:
@madsendennis Thanks a lot for all your efforts for removing vtk. I think with what you have provided so far we should bring us very close. I don't worry so much about the image part. We have a pure java nifti reader/writer. Image operations such as Gaussian smoothing are, as far as I know, not often used. We could just move them into a separate module. The module could, in essence just provide a converter between scalismo and vtk objects. In that way, a user could use all of vtk from Scalismo, without us having to invest a lot of effort in providing a tight integration.

madsendennis:
I discussed another possible move with @Andreas-Forster . 
To move everything (non-vtk) to a new library scalismo-core. Scalismo would then depend on this library + still include all the current VTK functionality. In this way, normal users will not see a difference in using scalismo, but we would then have a scalismo-core with a much smaller footprint, more appropriate for production use.
The main disadvantage as I see is that every update to scalismo-core will also trigger an update to scalismo and then scalismo-ui and scalismo-faces.

marcelluethi:
This is another good possibility. Maybe its even just a difference in naming. The difficulty with the transitive updates seems manageable. 

I have a preference for making Scalismo itself simpler rather than carving out the core. After the work you have now contributed, all essential functionality that are needed to write full shape modelling applications is there. If we name this Scalismo, we can tell users that they can get more functionality by combining it with the VTK library (at the expense of a couple of hundred MB footprint and reduced portability).  This feels to me different than offering a library called Scalismo, which has a footprint of 300 MB, but you can reduce it to a few MB by using only its core. I guess at this level it's just psychology.

[...]

PR #416, opened 2024-01-07
Feat/native mesh decimation

madsendennis:
This is a pure scala implementation of a mesh decimation algorithm based on the Fast Quadric Mesh Simplication algorithm: https://github.com/sp4cerat/Fast-Quadric-Mesh-Simplification/tree/master 

This is the first part of removing the VTK dependency in Scalismo, https://github.com/unibas-gravis/scalismo/issues/415


madsendennis:
The current version is a direct conversion of the original c++ version to a scala version, so a lot of code structure still needs to be properly updated. 
For now the algorithm works, so I would love some feedback on having this algorithm in comparison to the quadratic edge collapse, which might look more appealing. 
On a side note, this algorithm is with the current implementation already ~2 times faster (on the face mesh) than the VTK decimation we currently use. 

One minor bug is that some triangle norms are inverted.

Andreas-Forster:
I am fine with the replacement, if the normal inversion can be fixed. The old decimate could be offered as alternative in scalismo-vtk later on.

Andreas-Forster (changes requested):
I fine with the replacement except for the bug you mentioned.

- [x] Fix normal bug

Andreas-Forster (approved):

marcelluethi (approved):
I think the code can be merged. I would appreciate though, if you could run sbt scalafmtAll before merging, such that we have a consistent formatting in all of Scalismo.

madsendennis:
@Andreas-Forster is working on restructuring the code, I'll merge it in when we have a cleaner version.

PR #417, opened 2024-01-08
feat: remove vtk dependency to calculate tetrahedron volume

madsendennis:
Remove vtk dependency to calculate tetrahedron volume. 
Part of the steps to move away from VTK, https://github.com/unibas-gravis/scalismo/issues/415

### DIFF
--- a/src/main/scala/scalismo/mesh/TetrahedralMesh.scala
+++ b/src/main/scala/scalismo/mesh/TetrahedralMesh.scala
@@ -18,7 +18,7 @@ package scalismo.mesh
 import breeze.linalg.DenseVector
 import scalismo.common._
 import scalismo.geometry._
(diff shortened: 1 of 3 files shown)


Andreas-Forster:
I would vote for moving it to another location. @marcelluethi what do you think? Maybe `scalismo.numerics` would also be an option?

madsendennis:
> I would vote for moving it to another location. @marcelluethi what do you think? Maybe `scalismo.numerics` would also be an option?

Good idea. I think `scalismo.numerics` makes sense as well. I've pushed an update. Then hopefully it is directly ready to merge when Marcel gets time to look over it.

Andreas-Forster (approved):
Thank you for this PR.

marcelluethi (approved):
This looks good to me.

PR #420, opened 2024-01-16
Feat/remove vtk from scalismo

madsendennis:
Removes all VTK dependent functionality from Scalismo. This is mainly IO and format conversion. 

Upstream of this branch needs to be reset to main after the pending MRs are merged in. This branch contains the current version of the MRs. 

All vtk dependent tests have been moved to scalismo-vtk and commented out in this MR. 

Missing: The `gaussianSmoothing` has been commented out from `GaussianGradientImagePreprocessor`. We need to find another way to implement this in scalismo. 

By removing the VTK dependency, a simple scalismo.jar app goes from +700MB to ~40MB.

This is the last part of removing the VTK dependency from Scalismo, https://github.com/unibas-gravis/scalismo/issues/415


Andreas-Forster (changes requested):
I would definitively remove all commented code. If needed again, it is still in the history.

For the Gaussian image smoothing, it is not used so often. Maybe we can discuss to remove it temporarily at least and or  shift it also to the scalismo-vtk package.

- [x] Remove commented code

madsendennis:
Beside moving all VTK dependent functionality out of scalismo, we also move all active shape model code to a separate repository: https://github.com/unibas-gravis/scalismo-asm

madsendennis:
ASM removed from the code and outcommented tests. I already rebased it to be on top of the release-1.0 branch.

So I would say that it can be merged in - it does include the decimator that is still under rewriting and with a bug, bug we have a separate MR open for that. We should just not make a RC before that is fixed and tested.

madsendennis:
Branch rebased to on top of all the other MRs. Comments also removed from tests as requested by @Andreas-Forster .

Andreas-Forster (approved):

Issue #421, opened 2024-01-17
MAP estimate feature to avoid full posterior calculation


madsendennis:
To speed up applications that purely make use of `.posterior().mean`, it would be useful with a `.MAP()` alternative. 
This is both for the GP, DiscreteLowRank, PDM and SSM. 

In a quick test with a `PointDistributionModel[_3D, TriangleMesh]` with a 40 rank model with ~150.000 vertices and 1000 points to condition on, I get:
posterior.mean: ~900ms
MAP: ~90ms

For a pure GP `GaussianProcess[_3D, EuclideanVector[_3D]]`, the performance difference seems much less. Also with 1000 points to condition on, I get:
posterior.mean: ~6s
MAP: ~5.7s

madsendennis:
Merged in as part of scalismo 1.0

PR #422, opened 2024-01-17
Feat/gp map

madsendennis:
MR related to https://github.com/unibas-gravis/scalismo/issues/421

Provides handle to directly calculate the `.posterior.mean` with `.MAP` without calculating the posterior kernel/covariance matrix.

### DIFF
--- a/src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
+++ b/src/main/scala/scalismo/statisticalmodel/DiscreteLowRankGaussianProcess.scala
@@ -197,27 +197,25 @@ class DiscreteLowRankGaussianProcess[D: NDSpace, DDomain[DD] <: DiscreteDomain[D
 
   /**
    * Discrete version of
-   * [[DiscreteLowRankGaussianProcess.MAP(IndexedSeq[(Point[D], Vector[DO])], sigma2: Double]]. In contrast to
+   * [[DiscreteLowRankGaussianProcess.posteriorMean(IndexedSeq[(Point[D], Vector[DO])], sigma2: Double]]. In contrast to
    * this method, the points for the training data are defined by the pointId. The returned posterior process is defined
    * at the same points.
    */
-  def MAP(trainingData: IndexedSeq[(PointId, Value)],
-                sigma2: Double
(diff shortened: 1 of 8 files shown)


madsendennis:
I noted that in the `regression` function in `GaussianProcess.scala`, the inverse calculation is what takes up the majority of the time: `val invKernel: DenseMatrix[Double] = breeze.linalg.inv(K)`.
Does anyone have an idea how we can calculate this faster?

madsendennis:
I added an attempt to remove the inv from the mean and kernel calculation with the Cholesky decomposition. 
The tests pass, and the mean values seem to be the same in the instances I tested, but there seems to be some difference in the kernel calculation. Maybe @JonathanAellen can help have a look, especially on the last commit?

madsendennis:
I did some more test to check the difference for the inverse vs cholesky. They seem to provide the same output while cholesky is a lot faster to compute.

Andreas-Forster:
> I noted that in the `regression` function in `GaussianProcess.scala`, the inverse calculation is what takes up the majority of the time: `val invKernel: DenseMatrix[Double] = breeze.linalg.inv(K)`. Does anyone have an idea how we can calculate this faster?

Maybe you can have a look at using native libraries like blas under the hood (see https://github.com/luhenry/netlib)

Andreas-Forster:
For speed, there is a paper about cholesky and matrix inversion here: https://arxiv.org/abs/1111.4144

Andreas-Forster (approved):
I have more philosophical comments than real objection for the merge request:
- I do not like the name `MAP` as it is abbreviated and has an overlap with `map` from Scala.
- Further, I am not sure if we should make this functionality that accessible in our lib. ;-) Modeling probabilities, the MAP never tells you all you should want to know :-D

But as sait, I have no objection about merging it. It might be really usefull for some. Thanks for bringing that up.

marcelluethi:
@Andreas-Forster How about calling it posteriorMean instead of Map? This makes it clear that it is just a point estimate and also addresses the possible confusion with Scala's map function.

marcelluethi (approved):
Thanks for adding these methods. This looks like a useful addition to me.

PR #424, opened 2024-01-28
fix: search for triangle index update to work on mac

madsendennis:


madsendennis:
This "problem" that some ply formats cannot be read seems only to be a problem on Mac M processors when running a non-aarch64 java version. If changing to an aarch version, there is no problem.

madsendennis:
I am also able to replicate the problem on another machine (not arm processor) - this MR fixes it.

marcelluethi (approved):
Thanks for the fix. 
It is not obvious to me why one of the solution should work and the other not on some systems. But as the new solution is at the same time also more readable, I am happy if it is merged.

PR #425, opened 2024-02-01
Feature/modularize statismo io for better composition

Andreas-Forster:
This is a proposal for a change. Please comment or suggest freely.

Currently it is not possible to write two models into one hdf5file. One place where this is used is the ASM model. But having intensity models and volume models, it could also be interesting to write two models into one file.

This PR splits the functions into the file handling part and the actual writing of the data. Like this one could also open a file to write and write two models at different locations. The decision to make them private was to guard them somehow against accidental use due to IDE suggestion.

### DIFF
--- a/src/main/scala/scalismo/io/statisticalmodel/StatismoIO.scala
+++ b/src/main/scala/scalismo/io/statisticalmodel/StatismoIO.scala
@@ -82,10 +82,22 @@ object StatismoIO {
     canWarp: DomainWarp[D, DDomain],
     vectorizer: Vectorizer[EuclideanVector[D]]
   ): Try[PointDistributionModel[D, DDomain]] = {
-
-    val modelOrFailure = for {
+    for {
       h5file <- StatisticalModelIOUtils.openFileForReading(file)
+      model <- readStatismoPDM(h5file, modelPath)
+      _ <- Try(h5file.close())
+    } yield model
+  }
 
+  private[io] def readStatismoPDM[D: NDSpace, DDomain[D] <: DiscreteDomain[D]](
+    h5file: StatisticalModelReader,
+    modelPath: HDFPath
+  )(implicit
+    typeHelper: StatismoDomainIO[D, DDomain],
+    canWarp: DomainWarp[D, DDomain],
+    vectorizer: Vectorizer[EuclideanVector[D]]
+  ): Try[PointDistributionModel[D, DDomain]] = {
+    val modelOrFailure = for {
       mesh <- h5file.readStringAttribute(HDFPath(modelPath, "representer"), "datasetType") match {
         case Success("POINT_SET")    => readPointSetRepresentation(h5file, modelPath)
         case Success("POLYGON_MESH") => readStandardMeshRepresentation(h5file, modelPath)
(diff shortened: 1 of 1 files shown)


madsendennis:
Good idea. This would allow us to write a group of models to the same file, or just store different resolution versions of the same model in the same file, I see that this could be useful. 

I was wondering if it could be made more abstract such that each model could have a separate `[D, DDomain]`. Then the .5 could for instance contain different variants of the same model. Intensity, 2D/3D, tetrahedral, triangle mesh, low-res/high-res etc.

Andreas-Forster:
Note sure if I misunderstood something, but I think this is already possible. You just have to keep track of your model paths yourself at the moment. There would be the model catalogue which was maybe intended to do so somewhen, but it is never written anywhere.

I guess the following should work:
```scala
  val linePDM2d: PointDistributionModel[_2D, LineMesh] = ???
  val linePDM3d: PointDistributionModel[_3D, LineMesh] = ???
  val triangleModel3d: StatisticalMeshModel = ???
  val trianglePDM3d: PointDistributionModel[_3D, TriangleMesh] = PointDistributionModel(
    triangleModel3d.gp
  )
  val tetrahedralPDM3d: PointDistributionModel[_3D, TetrahedralMesh] = ???
  val imageIntensityModel3d: DiscreteLowRankGaussianProcess[_3D, DiscreteImageDomain, Float] = ???

  val file: File = ???
  Using(StatisticalModelIOUtils.openFileForWriting(file).get) { h5file =>
    for {
      _ <- StatismoIO.writeStatismoPDM(linePDM2d, h5file, HDFPath("/2d-line-model/"))
      _ <- StatismoIO.writeStatismoPDM(linePDM3d, h5file, HDFPath("/3d-line-model/"))
      _ <- StatismoIO.writeStatismoPDM(trianglePDM3d, h5file, HDFPath("/3d-mesh-model/"))
      _ <- StatismoIO.writeStatismoPDM(tetrahedralPDM3d, h5file, HDFPath("/3d-tetra-model/"))
      _ <- StatismoIO.writeStatismoImageModel(imageIntensityModel3d, h5file, HDFPath("/image-model/"))
    } yield ()
  }
```

Andreas-Forster:
We the recent commit, also an ASM (shape and appearance) can be written at a customizable path.

madsendennis:
For the example above, I was thinking if it would be possible to have a function like: 

```
def writeStatisticalShapeModels(models: Seq[(PointDistributionModel[D, Domain], path)], file: File)
```
And then inside the function match the type of model (D, Domain) with specific writers. 

Then in your example above you wouldn't need to manually open the file and specify the paths, but just create a Seq of all the models and apply those to the write function. 

But even the suggested update, I also like. I'm just throwing ideas around. :)

Andreas-Forster:
Thank you for the clarification. I got now your idea. Using this PR one could start experimenting around. As there are aside of PDMs also intensity and image models, I am not sure if we can easily provide a useful function that is not too generic or complex.

madsendennis:
That is true. In that case, I'm fine with merging in the MR as there is no change in the interface. Then it can be extended later if/when needed.

madsendennis (approved):

marcelluethi (approved):
Thanks for adding this feature. Would be happy to merge it.

PR #427, opened 2024-03-17
Fix/426 inconsitent reading and writing in statismoio

### LINKED ISSUE #332: SSM with no triangulation - read-write inconsistency
There is some inconsistency to the StatisticalModelIO when reading and writing an SSM without the reference mesh having a triangulation (empty TriangleList).
A file can successfully be written. But when read, it fails due to the processing of missing items in "representer/cells". 

Below is a small example which will show the error:

```scala
object ModelSaveLoadTest extends App {
  implicit val Random: Random = scalismo.utils.Random(1024)

  println("starting app...")
  scalismo.initialize()

  val zeroMean = VectorField(RealSpace[_3D], (_: Point[_3D]) => EuclideanVector.zeros[_3D])

  val cov: MatrixValuedPDKernel[_3D] = new MatrixValuedPDKernel[_3D]() {
    private val kernel = DiagonalKernel(GaussianKernel[_3D](1), 3)

    override protected def k(x: Point[_3D], y: Point[_3D]): DenseMatrix[Double] = {
      kernel(x, y)
    }
    override def outputDim = 3
    override def domain: Domain[_3D] = RealSpace[_3D]
  }

  val gp = GaussianProcess[_3D, EuclideanVector[_3D]](zeroMean, cov)

  val pointDomain = UnstructuredPointsDomain[_3D](IndexedSeq(Point3D(0, 0, 0), Point3D(1, 0, 0), Point3D(0, 1, 0)))
  val mesh = TriangleMesh3D(pointDomain, TriangleList(IndexedSeq()))

  val sampler = RandomMeshSampler3D(mesh, 10, 1024)
  val lowRankGP = LowRankGaussianProcess.approximateGPNystrom(gp, sampler, numBasisFunctions = 100)

  val ssm = StatisticalMeshModel(mesh, lowRankGP)

  val modelFile = new File("/tmp/dummy.h5")
  StatisticalModelIO.writeStatisticalMeshModel(ssm, modelFile)
  println("File written")
  StatisticalModelIO.readStatisticalMeshModel(modelFile).get
  println("File read")
}
```
Exception in thread "main" ncsa.hdf.hdf5lib.exceptions.HDF5Exception: ncsa.hdf.hdf5lib.exceptions.HDF5Exception: No data to read.

The error traces back to line 312 in the StatisticalModelIO.scala file.

### LINKED ISSUE #426: Inconsistent reading and writing in StatismoIO
There is an inconsistency in writing and reading data. This makes it possible to write models which can't be read in again. The concrete problem is: If no cells are present, nothing is written, but reading a file where no cells are written fails. The following lines show the issue:

Nothing is written if cells are empty here:
https://github.com/unibas-gravis/scalismo/blob/main/src/main/scala/scalismo/io/statisticalmodel/StatismoIO.scala#L162

Reading throws if cells are not present here:
https://github.com/unibas-gravis/scalismo/blob/main/src/main/scala/scalismo/io/statisticalmodel/StatismoIO.scala#L296
and `for`-comprehension does not handle the error but simply fails here:
https://github.com/unibas-gravis/scalismo/blob/main/src/main/scala/scalismo/io/statisticalmodel/StatismoIO.scala#L279

Proposal, we make sure that:
- the `for`-comprehension continues
- the cells are passed along as `None` when creating the domain.

Andreas-Forster:
This PR fixes #426 which is nescessary to write and read UnstructuredPointsDomains.

A test is introduced for writing and reading an UnstrucutedPointsDomain. The test fails without adapting the StatismoIO class. The test looks a bit more complicated, as only the private method is tested.

UPDATE:
This PR does also fixes #332.

Tests added which checks that passing None creates empty line or triangle lists.

### DIFF
--- a/src/main/scala/scalismo/io/statisticalmodel/StatismoDomainIO.scala
+++ b/src/main/scala/scalismo/io/statisticalmodel/StatismoDomainIO.scala
@@ -75,8 +75,8 @@ object StatismoDomainIO {
     override def createDomainWithCells(points: IndexedSeq[Point[_2D]],
                                        cellArray: Option[NDArray[Int]]
     ): Try[TriangleMesh[_2D]] = {
-      cellArray match {
-        case None => Failure(new Throwable("Triangle cells missing"))
+      val triangleList = cellArray match {
+        case None => Success(TriangleList(IndexedSeq()))
         case Some(c) =>
           val cellMatrix = ndIntArrayToIntMatrix(c)
           if (cellMatrix.cols != 3) Failure(new Exception("Representer cells are not triangles"))
@@ -84,10 +84,10 @@ object StatismoDomainIO {
             val cells = for (i <- 0 until cellMatrix.rows) yield {
               TriangleCell(PointId(cellMatrix(i, 0)), PointId(cellMatrix(i, 1)), PointId(cellMatrix(i, 2)))
             }
-            Success(TriangleMesh2D(UnstructuredPoints(points), TriangleList(cells)))
+            Success(TriangleList(cells))
           }
       }
-
+      triangleList.map(triangles => TriangleMesh2D(UnstructuredPoints(points), triangles))
     }
 
     override def cellsToArray(mesh: TriangleMesh[_2D]): NDArray[Int] = {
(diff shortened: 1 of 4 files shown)


Andreas-Forster:
As updated in the description, this PR is extended to alos fix #332.

marcelluethi (approved):
Thank you for fixing this. It looks great.

madsendennis (approved):
Looks good to me. Thanks for fixing this.

PR #433, opened 2025-07-30
Fix: Adjacent triangles for free points return empty list

### LINKED ISSUE #245: meshes with detached vertices
mesh.triangulation.adjacentTrianglesForPoint() currently results in a IndexOutOfBoundsException when accessed for a pointId not belonging to any cell (same for adjacentPointsForPoint). The reason is that the pointIds used for the computation of this array are obtained by iterating over the cells only.

This comes from the assumption that the meshes read into Scalismo must be "correct", i.e. without any floating vertices. This condition is however not enforced at read time.

Here we should either enforce that meshes are correct at read time (e.g. throw an exception and request to activate a repair flag) or fix the arrays above to return an empty Seq[TriangleId] for a point not belonging to any cell.

Andreas-Forster:
This PR fixes #245, which caused an `IndexOutOfBoundsException` when calling `adjacentTrianglesForPoint` on point IDs at the end of the vertex list that are not part of any triangle.

🛠️ Fix: The method now returns an empty list for:

- Point IDs that are not part of any triangle, and
- Point IDs that are not part of the triangle mesh at all

This ensures consistent and safe behavior for all inputs.

### DIFF
--- a/src/main/scala/scalismo/mesh/TriangleList.scala
+++ b/src/main/scala/scalismo/mesh/TriangleList.scala
@@ -42,7 +42,7 @@ case class TriangleList(triangles: IndexedSeq[TriangleCell]) {
     val dataSeq = IndexedSeq.tabulate(pointIds.size) { i =>
       data(pointIds(i)).toIndexedSeq
     }
-    id => dataSeq(id.id)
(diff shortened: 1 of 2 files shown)


Ghazi-Bouabene (approved):
Looks good. Thanks a lot.

PR #434, opened 2025-07-30
Fix: Clip triangle mesh keeps free floating vertices if they pass the predicate test

### LINKED ISSUE #311: Clip method in MeshOperations ignores points that are not part of a cell
When clipping a mesh with the method ```TriangleMesh3DOperations.clip```, one would assume that all the points of the mesh, which satisfy the given predicate are part of the resulting clipped mesh. However, the way it is currently implemented is, that only the points which satisfy the predicate and are part of a triangle are clipped. 

In Scalismo we do not enforce that every point in a mesh is part of a triangle. Indeed, it is perfectly valid in Scalismo to create a mesh with an empty triangle list. Therefore we should not assume or enforce that all points are part of a triangle in the clip mesh function. 

The same problem most likely occurs when clipping tetrahedral meshes.

Andreas-Forster:
This PR fixes #311 and keeps points which are not contained in any triangle in the mesh when they pass the test with the clip predicate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: 11
+      - uses: sbt/setup-sbt@v1
       - name: Run tests
         run:  |
           sudo apt-get update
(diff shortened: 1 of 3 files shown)


Ghazi-Bouabene (approved):
These changes are coherent with the current situation of allowing triangle meshes with floating vertices. So for me this looks good.

PR #435, opened 2025-07-31
Feature: Adds implementation to find n closest points in structured points

### LINKED ISSUE #57: DiscreteImageDomain.findNClosestPoints is not implemented

### LINKED ISSUE #397: Dropping the development branch and renaming master to main
When we started development of scalismo and scalismo-ui, we decided to follow the [git-flow](https://nvie.com/posts/a-successful-git-branching-model/) branching model. The main feature is to have a separate development branch. 

I would argue that at this point, this model is over-complicated and no longer needed. Instead I would propose to base development on the master branch. Furthermore, I propose to rename master to main, as this is the new standard everywhere. 

@Andreas-Forster You were also involved from the beginning. Do you see any reasons to keep the development branch?

Andreas-Forster:
This PR fixes #57 and fixes #397.

The implementation does a wave search starting from the closest point along the grid axis. This should be rather efficient for smaller N's while for very large N's one might to consider another implementation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: Scala CI
 
 on:
   push:
-    branches: [ master, release* ]
+    branches: [ main, release* ]
   pull_request:
-    branches: [ master, develop, release* ]
+    branches: [ main, release* ]
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


Ghazi-Bouabene (commented):
In many years of almost daily usage of Scalismo, I think I have never encountered this exception :) 

I would be in favor of skipping additional code if not needed. Not a strong opinion (can also understand the completeness argument).

PR #437, opened 2025-07-31
Fix: Make thread local variables consistent in their state in the surface spatial index

Andreas-Forster:
We observed that sometimes when using the `TriangleMeshInterpolator3D[EuclideanVector[_3D]]()`  in `approximateGPCholesky`  we had some artifacts, most of the time somewhere at the border.

This is due to a bug of non-consistent state in the two thread local variables `res` and `lastIdx` in the surface spatial index. This PR fixes this while still trying to preserve the deterministic behavior, which we wanted to establish with the changes that introduced the bug.


Ghazi-Bouabene (approved):
I can confirm, after testing the branch that the interpolation artifact at the border of the mesh disappears. Thanks a lot.

